PR #119, opened 2022-09-23
Rework collector tool

vstakhov:


vstakhov:
@sandreim I would suggest to treat as an initial tool refactoring. Then we would need to adopt the storage for things we have discussed: get rid of the `Weak` stuff, add multiple keys prefixes/namespaces (maybe even a hierarcy of multiple namespaces).

src/core/storage.rs
```diff
@@ -158,11 +162,20 @@ impl<K: Hash + Clone + Eq> RecordsStorage<K> {
 		self.ephemeral_records
 			.entry(block_number)
 			.or_insert_with(Default::default)
-			.insert(key, entry);
+			.insert(key, Arc::downgrade(&entry));
 
 		self.prune();
 	}
 
+	pub fn replace(&mut self, key: K, entry: StorageEntry) -> Option<StorageEntry> {
```
sandreim, line 170: We could merge this into the insert() method.

src/core/storage.rs
```diff
@@ -131,7 +135,7 @@ pub struct RecordsStorage<K: Hash + Clone> {
 	/// The last block number we've seen. Used to index the storage of all entries.
 	last_block: Option<BlockNumber>,
 	/// Elements with expire dates.
-	ephemeral_records: BTreeMap<BlockNumber, HashMap<K, Arc<StorageEntry>>>,
+	ephemeral_records: BTreeMap<BlockNumber, HashMap<K, Weak<StorageEntry>>>,
```
sandreim, line 138: Can you document the usage of `Weak` here please. I am not sure it is really needed.

sandreim (approved):

src/core/storage.rs
vstakhov: Please see the update.

src/core/storage.rs
vstakhov: Well, that might not be a good idea I think, as these methods have different signatures and meaning. Probably, we could use something like `entry` method for `HashMap` but it will introduce complexity and since we use storage for storing unique and distinct hashes, we have literally two different cases: 1) insert a new unique key (maybe even panic on duplicate) or 2) replace/update definitely or likely known key.

sandreim: Yeah, that makes sense only if we want to have uniqueness semantics for the write API.

sandreim (approved):
LGTM, merge when you think it's done.

PR #142, opened 2022-10-20
Add Grafana dashboards in JSON templates for KVDB

### LINKED ISSUE #106: Grafana dashboards: KVDB
JSON templates for monitoring relay chain and parachain db per column usage metrics.

bredamatt:
Closes #106 

- Adds JSON templates for Grafana Dashboards related to KVDB

Currently, the JSON model is exported directly from Grafana. We may evaluate the extent to which we want to autogenerate it in the future.


grafana/README.md
```diff
@@ -0,0 +1,9 @@
+# polkadot-introspector grafana dashboards
```
sandreim, line 1:
```suggestion
# Grafana dashboards
```

grafana/kvdb/README.md
```diff
@@ -0,0 +1,7 @@
+# KVDB Grafana dashboards
+
+This directory stores the Grafana dashboard JSON templates for the following introspector KVDB metrics:
```
sandreim, line 3: Why not have all of them in a single dashboard and remove the folder ?

sandreim (commented):
Added some comments. Please also link the grafana `README.md` into the top level one.

sandreim:
It would also be nice to provide a proper PR name instead of the default one.

grafana/kvdb/README.md
bredamatt: We should probably keep all the individual ones, and in stead add a combined one as well for optionality. Will add this later today.

grafana/kvdb/column-entry-count.json
```diff
@@ -0,0 +1,94 @@
+{
+  "id": 2,
+  "gridPos": {
+    "h": 8,
+    "w": 12,
+    "x": 0,
+    "y": 8
+  },
+  "type": "timeseries",
+  "title": "[${chain}] ${database} per column entry count",
```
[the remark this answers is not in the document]
bredamatt: Solved

grafana/kvdb/README.md
sandreim: I'm not sure it is worth exporting individual panels, or if it can actually work for the panels as currently defined. I also tried to import the PR JSONS and it doesn't work.

For things to work on import we also have to export the variables which are included in the `templating` JSON node.

grafana/kvdb/column-value-usage.json
```diff
@@ -0,0 +1,95 @@
...
+  "id": 1,
+  "gridPos": {
+    "h": 8,
+    "w": 12,
+    "x": 0,
+    "y": 0
+  },
+  "type": "timeseries",
+  "title": "[${chain}] ${database} per column value usage",
+  "description": "This panel shows the usage of each column value in the database.",
```
sandreim, line 11: I think think this can be better phrased so that it can be read as how much space all the values stored in a given column use.

grafana/kvdb/README.md
bredamatt: That helps! In that case, we should try with the entire JSON model. Will add it now (+ descriptions).

grafana/kvdb/column-value-usage.json
bredamatt: It is strange that they don't work since they are copied out from the previously working panel and copied into the repository. It could be that the `jq` formatting impacts how it works, but I would doubt it. Perhaps Grafana shows them differently when you expect the JSON panel relative to when you go directly to the JSON model in its entirety.

sandreim: The jq formatting is fine. The import succedes, but I get something unusable. I have to edit the queries and everything since the variables and their associated queries are gone. The only thing that should come from the user is the data source and the variable options should then be recomputed based on the queries.

sandreim: A picture is worth 1000 words. The variables are at the top of the screenshot: `source`, `chain`, `database`
Screenshot 2022-10-21 at 16 39 01

bredamatt: I can see the source, but I cannot see any of the chains, or the database options on the dashboard on my side. I am assuming these are defined in Thanos since that data source is there.

The query for chain also seems to be a bit misleading since it is defined as:

```
introspector_kvdb_values_size{}
```

Where can I find details on how to query the Thanos data source?

sandreim: Thanos is a Peometheus data source.

sandreim: There is a regexp IIRC that filters out the chain label value from that query.

bredamatt: Let me rephrase the above - Do we have any documentation explaining which queries we support? To me, the panels are not visible, so I cannot see what the correct queries are to make them appear as in your screenshot.

sandreim: https://grafana.parity-mgmt.parity.io/goto/b45roRNVz?orgId=1 - this is the original dashboard that you generated the JSONs from initially and it works for me, but it seems that the introspector deployment has been down for the past day, so you should select a timeframe before that in order for the queries to return data.

sandreim: > Let me rephrase the above - Do we have any documentation explaining which queries we support? To me, the panels are not visible, so I cannot see what the correct queries are to make them appear as in your screenshot.

Since this is a Prometheus data source, all Prometheus queries should work.

bredamatt: I managed to find the embedded queries in the link you shared above. Made sure to add these into the JSON model now. I changed the title of the dashboard to `KVDB Metrics` as well - it seemed more suitable.

bredamatt:
@sandreim  How much versi related documentation do you expect on this PR?

grafana/README.md
```diff
@@ -0,0 +1,26 @@
+# Grafana dashboards
+
+This directory holds different Grafana dashboard JSON models. For details about the Grafana dashboard JSON model, see: https://grafana.com/docs/grafana/latest/dashboards/build-dashboards/view-dashboard-json-model/
+
+In case you want to import these dashboards to Grafana, you can do so by following these steps:
+https://grafana.com/docs/grafana/v9.0/dashboards/export-import/. It is worthwhile mentioning that importing dashboards into Grafana requires Editor rights as a Grafana user.
+
+## Monitoring architecture
+
+Polkadot introspector is currently deployed in Versi. Versi is a test network which runs both a relay chain and several parachains in a Kubernetes environment.
```
sandreim, line 10: I don't think mentioning `Versi` is an important detail for users. We should just provide the steps for a successful deployment.

sandreim (approved):
LGTM!

vstakhov (approved):

PR #144, opened 2022-10-21
Fix clippy errors after toolchain upgrade

sandreim:


bredamatt (commented):
Out of curiosity, what was the motivation for moving the `SubxtCandidateEvent` to the heap and not the other `SubxtEvent` variants?

bredamatt (approved):

sandreim:
> Out of curiosity, what was the motivation for moving the `SubxtCandidateEvent` to the heap and not the other `SubxtEvent` variants?

https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant

PR #147, opened 2022-10-26
Implement prefixed and generic storage

vstakhov:
For some tools in the introspector, e.g. collector, we store multiple things using the same hash  (e.g. under relay chain block hash). To distinguish keys we need two changes:

* Use ordered structure (BTree) for keys, allowing to manipulate keys by some prefixes
* Change strict `H256` to a more generic type when dealing with the keys

### DIFF
--- a/src/collector/mod.rs
+++ b/src/collector/mod.rs
@@ -60,20 +60,26 @@ impl From<CollectorOptions> for WebSocketListenerConfig {
 	}
 }
 
+#[derive(Clone, Copy, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
+pub(crate) enum CollectorKeyPrefix {
+	Candidate,
+	Head,
+}
+
 /// Used to query storage for multiple hash types
 #[derive(Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
 pub(crate) struct CollectorKey {
-	pub prefix: String,
+	pub prefix: CollectorKeyPrefix,
 	pub hash: Option<H256>,
 }
 
 impl CollectorKey {
-	pub(crate) fn new_with_hash(prefix: &str, hash: H256) -> Self {
-		Self { prefix: prefix.into(), hash: Some(hash) }
+	pub(crate) fn new_with_hash(prefix: CollectorKeyPrefix, hash: H256) -> Self {
+		Self { prefix, hash: Some(hash) }
 	}
 
-	pub(crate) fn new_with_prefix(prefix: &str) -> Self {
-		Self { prefix: prefix.into(), hash: None }
+	pub(crate) fn new_with_prefix(prefix: CollectorKeyPrefix) -> Self {
+		Self { prefix, hash: None }
 	}
 }
 
@@ -82,9 +88,6 @@ impl HasPrefix for CollectorKey {
 		self.prefix == prefix.prefix
 	}
 }
-
-pub(crate) const CANDIDATE_PREFIX: &str = "candidate";
-pub(crate) const HEAD_PREFIX: &str = "head";
 
 pub(crate) async fn run(
 	opts: CollectorOptions,
@@ -170,7 +173,7 @@ async fn process_new_head(
 	api_service
 		.storage()
 		.storage_write(
-			CollectorKey::new_with_hash(HEAD_PREFIX, block_hash),
+			CollectorKey::new_with_hash(CollectorKeyPrefix::Head, block_hash),
 			StorageEntry::new_onchain(RecordTime::with_ts(header.number, Duration::from_secs(ts)), header.encode()),
 		)
 		.await
@@ -188,7 +191,7 @@ async fn process_candidate_change(
 		SubxtCandidateEventType::Backed => {
 			// Candidate should not exist in our storage
 			let maybe_existing = storage
-				.storage_read(CollectorKey::new_with_hash(CANDIDATE_PREFIX, change_event.candidate_hash))
+				.storage_read(CollectorKey::new_with_hash(CollectorKeyPrefix::Candidate, change_event.candidate_hash))
 				.await;
 			if let Some(existing) = maybe_existing {
 				let candidate_descriptor: CandidateRecord = existing.into_inner()?;
@@ -201,7 +204,7 @@ async fn process_candidate_change(
 				// Find the relay parent
 				let maybe_relay_parent = storage
 					.storage_read(CollectorKey::new_with_hash(
-						HEAD_PREFIX,
+						CollectorKeyPrefix::Head,
 						change_event.candidate_descriptor.relay_parent,
 					))
 					.await;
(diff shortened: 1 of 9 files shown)


src/collector/mod.rs
```diff
@@ -60,6 +60,16 @@ impl From<CollectorOptions> for WebSocketListenerConfig {
 	}
 }
 
+/// Used to query storage for multiple hash types
+#[derive(Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
+pub(crate) struct CollectorKey {
+	pub prefix: String,
+	pub hash: Option<H256>,
```
sandreim, line 67: why is this optional ?

src/core/api/storage.rs
```diff
@@ -20,46 +20,46 @@ use crate::{
...
-	Write(H256, StorageEntry),
-	Read(H256),
-	Replace(H256, StorageEntry),
+pub enum RequestType<K: Ord + Hash + Sized> {
+	Write(K, StorageEntry),
+	Read(K),
+	Replace(K, StorageEntry),
 	Size,
-	Keys,
+	Keys(Option<K>),
```
sandreim, line 36: It would be more explicit to have a variant `KeysWithPrefix` which also gets rid of the `Option`

src/core/storage.rs
```diff
@@ -225,6 +225,15 @@ impl<K: Hash + Clone + Eq + Debug> RecordsStorage<K> {
...
 		self.direct_records.keys().cloned().collect()
 	}
+
+	/// Returns keys starting from a specific prefix (for string keys)
+	pub fn keys_prefix<Q: Sized>(&self, prefix: &Q) -> Vec<K>
+	where
+		K: Borrow<Q>,
+		Q: Ord,
+	{
+		self.direct_records.range(prefix..).map(|(k, _)| k).cloned().collect()
```
sandreim, line 235: DQ: Does this really work ?

src/collector/mod.rs
vstakhov: For prefixed searches, as `None < Some(H256)`.

src/core/api/storage.rs
vstakhov: Yes, I was thinking about it, good point.

src/core/storage.rs
vstakhov: I'm currently writing a test for it :)

src/collector/mod.rs
```diff
@@ -60,6 +61,70 @@ impl From<CollectorOptions> for WebSocketListenerConfig {
...
+
+impl CollectorKey {
+	pub(crate) fn new_relay_chain_block(hash: H256) -> Self {
+		Self { prefix: CollectorPrefixType::Head, hash }
+	}
+	pub(crate) fn new_parachain_candidate(hash: H256, para_id: u32) -> Self {
+		Self { prefix: CollectorPrefixType::Candidate(Some(para_id)), hash }
+	}
+	// This is used when we don't know the parachain id
+	pub(crate) fn new_generic_hash(hash: H256) -> Self {
```
sandreim, line 90: Do we really need this ? I think we can drop the Option from `CollectorPrefixType::Candidate`

src/collector/mod.rs
```diff
@@ -60,6 +61,70 @@ impl From<CollectorOptions> for WebSocketListenerConfig {
...
+/// Used to query storage for multiple hash types
+#[derive(Clone, Debug, Ord, PartialOrd, Eq)]
+pub(crate) struct CollectorKey {
+	/// Prefix for a specific collector key
+	pub prefix: CollectorPrefixType,
+	/// Block or a candidate hash
+	pub hash: H256,
+}
+
+impl CollectorKey {
```
sandreim, line 82: Since all members are public it is so easy to construct it, so we could not implement these at all.

src/collector/mod.rs
```diff
@@ -60,6 +61,70 @@ impl From<CollectorOptions> for WebSocketListenerConfig {
...
+		Self { prefix: CollectorPrefixType::Candidate(None), hash }
+	}
+
+	pub(crate) fn new_with_prefix(prefix: CollectorPrefixType) -> Self {
+		Self { prefix, hash: H256::zero() }
+	}
+}
+
+impl HasPrefix for CollectorKey {
+	fn has_prefix(&self, other: &Self) -> bool {
```
sandreim, line 100: Does this actually check for the same prefix ? It is a bit confusing, the trait definition says:
```
	/// Returns whether there is a specified prefix at the start
	/// The default implementation assumes that there is no prefix at all
```

src/collector/mod.rs
```diff
@@ -60,6 +61,70 @@ impl From<CollectorOptions> for WebSocketListenerConfig {
...
+		}
+	}
+}
+
+// This implementation compares the hashes and optionally checks for the prefix
+// For example, when we have no parachain id available, we should be able to find
+// a specific entry with no hash
+impl PartialEq for CollectorKey {
+	fn eq(&self, other: &Self) -> bool {
+		self.hash == other.hash && (self.has_prefix(other) || other.has_prefix(self))
```
sandreim, line 118: Isn't `self.has_prefix(other)` equal to `other.has_prefix(self)` ?

src/collector/mod.rs
sandreim: I am not sure we even need this trait if we consider it just as the first part of the key blob. This logic would just be internal to the storage module.

src/collector/mod.rs
vstakhov: We don't know parachain id in the dispute events... There is also a convenient method to get information about a candidate by just a hash, so this function should be implemented I think (lookup without para id I mean).

src/collector/mod.rs
vstakhov: Yes, that was the initial implementation, but I've decided that having convenience create methods is good from the terms of style.

src/collector/mod.rs
vstakhov: Well, I have another thought, that we should actually have a storage that implements `keys_prefix` when the key type T satisfies `HasPrefix` trait. It might be a more convenient interface. With regard to the first comment, I think this should be a part of another PR to fix collector.

src/collector/mod.rs
vstakhov: It is now, but it wasn't when I've experimented with this code. As previously, that might be better to move this to a separate PR.

src/collector/mod.rs
vstakhov: I have implemented this approach. So the storage is normal by default but can optionally be created with prefixes support when `K` has `HasPrefix` trait. It will leave the remaining tools untouched (and there no longer a requirement to implement `HasPrefix` for untouched types) and allows collector to do prefixed stuff.

vstakhov:
Ok, I actually found that this idea is also non-viable, and the easiest way to do it to have multiple struct types for storage and prefixed storage and nested hash maps.

PR #149, opened 2022-11-07
Another try to implement prefixed storage and fix the collector

vstakhov:
This PR is intended to fix the following tasks:

- [x] Allow to have generic storage
- [x] Allow to have a specific storage version to store hashes with prefixes
- [x] Do not break other tools
- [x] Implement prefixed and non-prefixed methods for storage types (prefixed storage is an extended generic storage)
- [x] Fix collector tool to work with prefixes properly
- [x] Adjust tests
- [x] Fix websocket counterpart to support prefixed searches
- [x] Allow to get prefixes from a storage

### DIFF
--- a/src/core/storage.rs
+++ b/src/core/storage.rs
@@ -27,7 +27,7 @@ use crate::eyre;
 use codec::{Decode, Encode};
 use std::{
 	borrow::Borrow,
-	collections::{HashMap, HashSet},
+	collections::{BTreeMap, HashMap, HashSet},
 	fmt::Debug,
 	hash::Hash,
 	time::Duration,
@@ -166,7 +166,7 @@ pub struct HashedPlainRecordsStorage<K: Hash + Clone> {
 	/// The last block number we've seen. Used to index the storage of all entries.
 	last_block: Option<BlockNumber>,
 	/// Elements with expire dates.
-	ephemeral_records: HashMap<BlockNumber, HashSet<K>>,
+	ephemeral_records: BTreeMap<BlockNumber, HashSet<K>>,
 	/// Direct mapping to values.
 	direct_records: HashMap<K, StorageEntry>,
 }
@@ -176,7 +176,7 @@ where
 	K: Hash + Clone + Eq + Debug,
 {
 	fn new(config: RecordsStorageConfig) -> Self {
-		let ephemeral_records = HashMap::new();
+		let ephemeral_records = BTreeMap::new();
 		let direct_records = HashMap::new();
 		Self { config, last_block: None, ephemeral_records, direct_records }
 	}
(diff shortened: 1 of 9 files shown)


src/core/storage.rs
```diff
@@ -206,27 +230,213 @@ impl<K: Hash + Clone + Eq + Debug> RecordsStorage<K> {
...
 	}
 }
 
+/// This trait is used to define a storage that can store items organised in prefixes.
+/// Prefixes are used to group elements by some characteristic. For example, to get
+/// elements that belong to some particular parachain.
+pub trait PrefixedRecordsStorage<K, P> {
+	/// Insert a prefixed entry to the storage
+	fn insert_prefix(&mut self, prefix: P, key: K, entry: StorageEntry) -> color_eyre::Result<()>;
+	/// Replaces a prefixed entry in the storage, both prefix and a key must exist
```
sandreim, line 256: What does it return ?

src/core/storage.rs
```diff
@@ -133,10 +133,34 @@ pub struct RecordsStorageConfig {
 	pub max_blocks: usize,
 }
 
+/// This trait defines basic functions for the storage
+pub trait RecordsStorage<K> {
+	/// Creates a new storage with the specified config
+	fn new(config: RecordsStorageConfig) -> Self;
+	/// Inserts a record in ephemeral storage. This method does not overwrite
+	/// records and returns an error in case of a duplicate entry.
+	fn insert(&mut self, key: K, entry: StorageEntry) -> color_eyre::Result<()>;
```
sandreim, line 142: what happens if the key already exists ?

src/core/storage.rs
```diff
@@ -206,27 +230,213 @@ impl<K: Hash + Clone + Eq + Debug> RecordsStorage<K> {
...
+{
+	fn new(config: RecordsStorageConfig) -> Self {
+		let ephemeral_records = BTreeMap::new();
+		let prefixed_records = HashMap::new();
+		Self { config, last_block: None, ephemeral_records, prefixed_records }
+	}
+
+	// We cannot insert non prefixed key into a prefixed storage
+	fn insert(&mut self, key: K, _: StorageEntry) -> color_eyre::Result<()> {
+		return Err(eyre!("trying to insert key with no prefix to the prefixed storage: {:?}", key))
```
sandreim, line 310: We could also also allow this operation and use a default prefix. It is more flexible and would allow old written code to still work with a prefixed storage, even if it's code doesn't actually need prefixes.

sandreim (approved):
Nice!

I have one overall nit regarding the presence of 4 extra APIs Rad/write/replace/keys. I would not have exposed them at all, but create 2 macros for constructing plain and prefixed keys via concatenation plus making Keys take an optional `prefix` param. Your approach seems to favor performance by using a 2nd level hashmap for prefixes allowing faster lookup times. I would also make this tradeoff only if were would be constrained by performance.

src/core/storage.rs
vstakhov: Same as insert: `Ok()` on insertion or an error on duplicate entry.

src/core/storage.rs
vstakhov: An error is returned, as we discussed, the keys must be distinct, so a duplicate is likely an error in the logic of the code.

src/core/storage.rs
vstakhov: One problem I've found is to define what is a default prefix for type `P`? We could use `Default` trait for that but it adds an additional burden on `P` type...

PR #166, opened 2022-11-25
Support three Runtimes at build time using features

sandreim:
The docker image will contain both builds but default to polkadot as an entrypoint.


sandreim:
Also added Versi, in practice they might be different but we control it easily as it is private. There are a few tests that run against Polkadot and Rococo rpc nodes which are public, but not doing the same for Versi.

vstakhov:
We should really think about including multiple metadatas at the compile time. However, this is definitely a better option than we have so far!

vstakhov (approved):

sandreim:
This is just the path of least resistance to fix the annoyance of having to do code changes to switch between runtimes. It would be really cool to solve this with a single build, but so far I didnt get to think deep about it.

Issue #112, opened 2022-09-14
Display summary at exit


sandreim:
This should be available for all cli tools, pretty much how icmp ping stats look at exit. Some examples
- event count per type
- Avg, min max block time
- Disputed candidate count
- Dispute votes count per block, total, etc
- 

Additionally it could make sense to also have this info dumped in json format for easy integration with other tooling.

PR #167, opened 2022-11-25
Run for a number of blocks

sandreim:
Small chore before implementing #112 for parachain commander.


vstakhov:
I was about to suggest using a signed number to avoid too many `Option` handling, but it seems we have already `Option` based interface there, so my suggestion won't be sound.

vstakhov (approved):

PR #181, opened 2022-12-14
Implement subscribe mode feature

### LINKED ISSUE #180: Parachain Commander: handle forks
Stats might be incorrect when we have forks (we subscribe to the best block stream). This should be handled and accounted for.

```
2022-12-14T07:28:36.000000000Z +6000ms [#63168]
        CANDIDATE BACKED
        💜 Candidate hash: 0xd4369e2e5a4ad4c541c23790d6849497d99b12f85f89a956b5d3d5bd00131280 
        🔗 Relay block hash: 0xe1b285daefe9c9e84bb91a2211fd69771eff7742bc3c8a875597078eb0abc66f 
        🥝 Availability core OCCUPIED

2022-12-14T07:28:36.001000000Z +1ms [#63168]
        CANDIDATE BACKED
        💜 Candidate hash: 0xd4369e2e5a4ad4c541c23790d6849497d99b12f85f89a956b5d3d5bd00131280 
        🔗 Relay block hash: 0x5b7c92b2ba448f305150564cbbc6894bcce3015614140c18e23b6d8aafe469ff 
        🥝 Availability core OCCUPIED
```

```
2022-12-14T07:29:42.000000000Z +6000ms [#63179]
        CANDIDATE INCLUDED
        💜 Candidate hash: 0xceb89b2b7dcb71f6b8f5d5e800f399e7f81f9ec99a8b94b4ebd5f29027485508 
        🟢 Availability bits: 259/260
        🔗 Relay block hash: 0xb64f33265bcfb17054207e511a22c2c9a07eaa971ebee8786995413b015add48 
        🥝 Availability core FREE

2022-12-14T07:29:42.001000000Z +1ms [#63179]
        SLOW BACKING, no candidate backed
        🔗 Relay block hash: 0x65a3c5088a051111213916d4d04448c77dad5af4f1675fe9f66bd58e170536de 
        🥝 Availability core OCCUPIED
```

vstakhov:
This PR implements different subscribe modes:

```
  -s, --subscribe-mode <SUBSCRIBE_MODE>
          Defines subscription mode

          [default: subscribe-all]

          Possible values:
          - subscribe-all:       Subscribe to all blocks
          - subscribe-best:      Subscribe to the best chain
          - subscribe-finalized: Subscribe to finalized blocks
```

Closes #180

### DIFF
--- a/src/collector/mod.rs
+++ b/src/collector/mod.rs
@@ -33,7 +33,7 @@ mod ws;
 
 use crate::core::{
 	ApiService, EventConsumerInit, RecordTime, RecordsStorageConfig, StorageEntry, StorageInfo, SubxtCandidateEvent,
-	SubxtCandidateEventType, SubxtDispute, SubxtDisputeResult, SubxtEvent,
+	SubxtCandidateEventType, SubxtDispute, SubxtDisputeResult, SubxtEvent, SubxtSubscriptionMode,
 };
 use candidate_record::*;
 use color_eyre::eyre::eyre;
@@ -52,6 +52,9 @@ pub(crate) struct CollectorOptions {
 	/// WS listen address to bind to
 	#[clap(short = 'l', long = "listen", default_value = "0.0.0.0:3030")]
 	listen_addr: SocketAddr,
+	/// Defines subscription mode
+	#[clap(short = 's', long = "subscribe-mode", default_value_t, value_enum)]
+	pub subscribe_mode: SubxtSubscriptionMode,
 }
 
 impl From<CollectorOptions> for WebSocketListenerConfig {
--- a/src/collector/mod.rs
+++ b/src/collector/mod.rs
@@ -53,7 +53,7 @@ pub(crate) struct CollectorOptions {
 	#[clap(short = 'l', long = "listen", default_value = "0.0.0.0:3030")]
 	listen_addr: SocketAddr,
 	/// Defines subscription mode
-	#[clap(short = 's', long = "subscribe-mode", default_value_t, value_enum)]
+	#[clap(short = 's', long = "subscribe-to", default_value_t, value_enum)]
 	pub subscribe_mode: SubxtSubscriptionMode,
 }
 
--- a/src/core/subxt_subscription.rs
+++ b/src/core/subxt_subscription.rs
@@ -225,7 +225,7 @@ impl SubxtWrapper {
 // Subxt does not export this type but it is needed to specify future output
 type BlockStream<T> = Pin<Box<dyn Stream<Item = Result<T, Error>> + Send>>;
 
-// Process subscription to a specific block types (e.g. all, best, finalized) and returns `true`
+// Process subscription to a specific block types (e.g. best, finalized) and returns `true`
 // if a caller's loop should be terminated.
 async fn process_subscription_or_stop<Sub, Client>(
 	update_channel: &Sender<SubxtEvent>,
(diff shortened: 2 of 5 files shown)


src/collector/mod.rs
```diff
@@ -52,6 +52,9 @@ pub(crate) struct CollectorOptions {
 	/// WS listen address to bind to
 	#[clap(short = 'l', long = "listen", default_value = "0.0.0.0:3030")]
 	listen_addr: SocketAddr,
+	/// Defines subscription mode
+	#[clap(short = 's', long = "subscribe-mode", default_value_t, value_enum)]
```
sandreim, line 56:
```suggestion
	#[clap(short = 's', long = "subscribe-to", default_value_t, value_enum)]
```

src/core/subxt_subscription.rs
```diff
@@ -69,6 +73,23 @@ pub enum SubxtDisputeResult {
...
 
+/// How to subscribe to subxt blocks
+#[derive(strum::Display, Debug, Clone, Copy, clap::ValueEnum)]
+pub enum SubxtSubscriptionMode {
+	/// Subscribe to all blocks
+	SubscribeAll,
+	/// Subscribe to the best chain
+	SubscribeBest,
+	/// Subscribe to finalized blocks
+	SubscribeFinalized,
```
sandreim, line 84:
```suggestion
pub enum SubxtSubscriptionMode {
	/// Subscribe to all blocks
	All,
	/// Subscribe to the best chain
	Best,
	/// Subscribe to finalized blocks
	Finalized,
```

sandreim (commented):
Looks good, but I don't think it addresses the para block time stats, since we don't care which fork is chosen. I've seen cases where one fork had a candidate backed and another that didnt.

vstakhov:
> Looks good, but I don't think it addresses the para block time stats, since we don't care which fork is chosen. I've seen cases where one fork had a candidate backed and another that didnt.

Do you mean block time command?

sandreim:
The parachain commander para block times we show in summary

sandreim:
I am unsure we need the subscribe all, that would need special handling. Just finalized/best is enough

vstakhov:
Then it is probably better to change the default to `subscribe_best` and then leave `subscribe_all` for special use cases.

src/core/subxt_subscription.rs
```diff
@@ -69,6 +73,23 @@ pub enum SubxtDisputeResult {
 	TimedOut,
 }
 
+/// How to subscribe to subxt blocks
+#[derive(strum::Display, Debug, Clone, Copy, clap::ValueEnum)]
+pub enum SubxtSubscriptionMode {
+	/// Subscribe to all blocks
+	All,
```
sandreim, line 80: If we choose this do we get notifications twice for a block, when it's first imported and then finalized ? How do the tools cope with this ?

vstakhov: I have not tested this I'm afraid. Presumably, it will be broken (unless my next WiP PR about collector-as-a-service is finished). However, without this PR the tool is definitely broken as it uses `subscribe_all` by default.

sandreim: Let’s just not expose the subscribe all here so we don’t break things at all

vstakhov: Done.

sandreim (approved):
Looks good!

Issue #179, opened 2022-12-13
Dispute tracker: use events 


sandreim:
Currently we count votes ourselves, we could make this more robust by using the `DisputeConcluded` events generated by the runtime logic.

vstakhov:
It is done for now: events are used as the real source of disputes outcomes.

PR #187, opened 2022-12-19
Use runtime events to filter dispute tracking

vstakhov:
This PR addresses #179. Aside of the function listed in the issue, this approach also allows to track time between disputes initiation and conclusion, for example:

```
Disputes stats: 259 disputes tracked, 259 concluded valid, 0 concluded invalid, 1.88 blocks average resolution time, 1.6 average misbehaving validators
```


src/pc/tracker.rs
```diff
@@ -330,18 +350,44 @@ impl ParachainBlockTracker for SubxtTracker {
...
+
+	fn inject_disputes_events(
+		&mut self,
+		disputes_tracked: &HashMap<Self::RelayChainBlockHash, (Self::DisputeOutcome, Option<u32>)>,
+	) {
+		self.disputes.retain_mut(|dispute| -> bool {
+			if let Some(observed_dispute) = disputes_tracked.get(&dispute.candidate) {
+				if dispute.outcome != observed_dispute.0 {
+					info!(
+						"Dispute for candidate {:?}: calculated outcome: {:?} is not equal to observed outcome: {:?}",
```
sandreim, line 368: When/How can this happen ?

src/pc/tracker.rs
```diff
@@ -473,7 +519,10 @@ impl SubxtTracker {
 					.count() as u32;
 				let voted_against = dispute_info.statements.len() as u32 - voted_for;
 
-				let misbehaving_validators = if voted_for > voted_against {
+				let outcome =
+					if voted_for > voted_against { SubxtDisputeResult::Valid } else { SubxtDisputeResult::Invalid };
```
sandreim, line 523: shouldn't the outcome be  based on events ?

src/pc/tracker.rs
```diff
@@ -70,16 +82,20 @@ struct SubxtSessionTracker {
...
 	pub candidate: H256,
+	/// The real outcome
+	pub outcome: SubxtDisputeResult,
 	/// Number of validators voted that a candidate is valid
 	pub voted_for: u32,
 	/// Number of validators voted that a candidate is invalid
 	pub voted_against: u32,
 	/// A vector of validators voted against supermajority (index + identify)
 	pub misbehaving_validators: Vec<(u32, String)>,
+	/// How many blocks have passed since DisputeInitiated event
```
sandreim, line 97:
```suggestion
	/// Dispute conclusion time: how many blocks have passed since DisputeInitiated event
```

src/pc/tracker.rs
```diff
@@ -70,16 +82,20 @@ struct SubxtSessionTracker {
 }
 
 /// An outcome for a dispute
-#[derive(Encode, Decode, Debug, Default, Clone)]
+#[derive(Encode, Decode, Debug, Clone)]
 pub struct DisputesOutcome {
```
sandreim, line 86: I would rename this struct, as we have `the real outcome` inside of it. `DisputeProgress` or `DisputeInfo` ?

src/pc/tracker.rs
vstakhov: At this point we might not see the corresponding event, as it comes later.

sandreim: I think we should only take the outcome from the events when it comes later and not rely on these calculations.

vstakhov: The main problem is that the result from the events comes later, so at that point we cannot find misbehaving validators, which is a useful functionality. In fact, this fix should be postponed till we have the proper collector service. In that case, we can gather disputes information for a single candidates from multiple inherent datas and join it together. For now, I see no other way to preserve partial data between multiple blocks. And I'm inclined not to remove this code completely as it is very useful to find misbehaving validators. Even if it is not now 100% correct, it is better to fix it once we can keep and share data across multiple blocks.

sandreim: Gotcha, makes sense.

sandreim (approved):

PR #202, opened 2023-01-10
Implement collector as a service

vstakhov:
The collector tool should be converted to a service to track blocks over parachains. Here is a list of the tasks:

- [x] Move collector to a service
- [x] Add handling for the subxt events
- [x] Unify graceful shutdown
- [x] Propagate websocket interface
- [x] Implement necessary storage support for persistent records
- [x] Implement storage access methods
- [x] Convert tools to use collector storage where needed
- [x] Support for the relay chain forks:
  * [x] Do not mess the average relay time when forks are found
  * [ ] Do not indicate slow availability/missed slots if one fork has candidates and another doesn't
  * [ ] Do not mess the metrics (maybe we should just use finalized chain for that purposes)
- [ ] Test and (possibly) fix disputes display and metrics


src/core/collector/candidate_record.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Parity Technologies (UK) Ltd.
+// Copyright 2023 Parity Technologies (UK) Ltd.
```
sandreim, line 1:
```suggestion
// Copyright 2022-2023 Parity Technologies (UK) Ltd.
```

src/core/storage.rs
```diff
@@ -42,6 +42,8 @@ pub enum RecordSource {
 	Onchain,
 	/// For offchain data.
 	Offchain,
+	/// For persistent data (not pruned automatically)
+	Persistent,
```
sandreim, line 46: What do we want to use this for ? Also, I don't think this is a good place for it. This represents the source of the record, not its lifetime.

vstakhov: We want this for session information, that should not be prunned. However, placement it in the RecordSource is really a logical mistake. So we can either tell that it is not a `RecordSource` but `RecordType` or add an additional enum to distinguish block based and persistent data.

src/core/collector/candidate_record.rs
vstakhov: In this case, it should be done globally in a separate commit/PR I suppose.

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+pub enum CollectorPrefixType {
+	/// Candidate prefixed by Parachain-Id
+	Candidate(u32),
+	/// A mapping to find out parachain id by candidate hash (e.g. when we don't know the parachain)
+	CandidatesParachains,
+	/// Relay chain block
+	Head,
+	/// Session data, keyed by session hash (blake2b(session_index))
+	Session,
+	/// Inherent data (more expensive to store, so good to have it shared)
```
sandreim, line 75: I think we have to be careful here and adjust the prune settings. Right now we keep stuff for 1000 blocks, so we end up with high memory requirements for running in Prometheus mode without actually needing it.

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+struct CollectorState {
+	/// The current block number
+	current_relay_chain_block_number: u32,
+	/// A list of hashes at this height (e.g. if we see forks)
+	current_relay_chain_block_hashes: Vec<H256>,
+	/// A list of candidates seen, indexed by parachain id
+	candidates_seen: BTreeMap<u32, Vec<H256>>,
+	/// A list of disputes seen, indexed by parachain id
+	disputes_seen: BTreeMap<u32, Vec<DisputeInfo>>,
+	/// A current session index
```
sandreim, line 106:
```suggestion
	/// The current session index
```

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+	current_session_index: u32,
+}
+
+/// Provides collector new head events split by parachain
+#[derive(Clone, Debug)]
+pub struct NewHeadEvent {
+	/// Relay parent block number
+	pub relay_parent_number: <PolkadotConfig as subxt::Config>::BlockNumber,
+	/// Relay parent block hash (or hashes in case of the forks)
+	pub relay_parent_hashes: Vec<H256>,
```
sandreim, line 116: I think this shouldn't be a vec as I would expect to see distinct `NewHeadEvent` with same block number.

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+	/// The target parachain id (used for broadcasting events)
+	pub para_id: u32,
+	/// Candidates seen for this relay chain block that belong to the target parachain
```
sandreim, line 119: What does target mean in this context?

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+	pub candidates_seen: Vec<H256>,
+	/// Disputes concluded in these blocks
```
sandreim, line 121:
```suggestion
	/// Disputes concluded in this block
```

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+			"imported new block hash: {:?}, number: {}, previous number: {}, previous hashes: {:?}",
+			block_hash,
+			block_number,
+			self.state.current_relay_chain_block_number,
+			self.state.current_relay_chain_block_hashes
+		);
+
+		match block_number.cmp(&self.state.current_relay_chain_block_number) {
+			Ordering::Greater => {
+				self.update_state(block_number, block_hash)?;
```
sandreim, line 318: To avoid any possible racing vs consumers, update must to be sent only after we write everything related to this block in storage.

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+	/// Occurs when collector is disconnected and is about to terminate
+	Termination,
+}
+
+pub struct Collector {
+	api_service: CollectorStorageApi,
+	ws_listener: Option<WebSocketListener>,
+	to_websocket: Option<BroadcastSender<WebSocketUpdateEvent>>,
+	endpoint: String,
+	subscribe_channels: BTreeMap<u32, Vec<BroadcastSender<CollectorUpdateEvent>>>,
```
sandreim, line 141: I am wondering why we need both subscriptions and broadcast channels, for simplicity we could only use broadcast and filter by para id at consumer level.

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+	broadcast_channels: Vec<BroadcastSender<CollectorUpdateEvent>>,
+	state: CollectorState,
+}
+
+impl Collector {
+	pub fn new(endpoint: &str, opts: CollectorOptions) -> Self {
+		let api_service: CollectorStorageApi =
+			ApiService::new_with_prefixed_storage(RecordsStorageConfig { max_blocks: opts.max_blocks.unwrap_or(1000) });
```
sandreim, line 149: I'd make this 2 digit default. I don't know any usecase that can benefit from a high default right now.

sandreim (commented):
Done a first pass. A few questions, suggestions and nits.

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+				.storage_write_prefixed(
+					CollectorPrefixType::Session,
+					cur_session_hash,
+					StorageEntry::new_persistent(
+						RecordTime::with_ts(block_number, Duration::from_secs(ts)),
+						accounts_keys,
+					),
+				)
+				.await?;
+			// Remove old session with the index `cur_session - 2` ignoring possible errors
```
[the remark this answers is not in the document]
vstakhov: In fact, we need only the current and the previous session indeed. I was thinking about outdated disputes.

src/core/collector/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+	listen_addr: Option<SocketAddr>,
+}
+
+/// This type is used to distinguish different keys in the storage
+#[derive(Clone, Copy, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
+pub enum CollectorPrefixType {
+	/// Candidate prefixed by Parachain-Id
+	Candidate(u32),
+	/// A mapping to find out parachain id by candidate hash (e.g. when we don't know the parachain)
+	CandidatesParachains,
```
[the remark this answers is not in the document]
vstakhov: Well, we actually can search a candidate hash without a prefix, I did this mapping for some sanity checks (and somehow for the performance). Dispute events have no information about parachain, and that's why we still need that mapping at some implementation (either by checking each prefix or by having an additional mapping).

src/core/collector/mod.rs
vstakhov: Updates for the current block are sent when we receive a block with higher height. 

I don't know if we can receive events that belong to a block with lower height after we have seen a block with some current height that is more than that. It doesn't seem possible on a finalized chain at least.

src/core/collector/mod.rs
vstakhov: It is now 64, thanks!

src/core/collector/mod.rs
vstakhov: It is now 64 blocks by default, so it shouldn't be a big issue (under 100M of memory I suppose).

src/core/collector/mod.rs
vstakhov: Mainly for the convenience. Per-parachain subscriptions are the only subscriptions that exist for now. We might want broadcast subscriptions at some point as well. Filtering on the consumer side is possible but might be noisy, as we normally need a single parachain but will receive quite a lot of traffic about unrelated events.

sandreim (approved):

Issue #193, opened 2023-01-03
Parachain Commander: Allow to track multiple parachains (or even all parachains)


vstakhov:
This issue relies on #192 and also implies that we store some data locally (e.g. inherent data) to avoid RPC nodes overload. Display part is not very important and Prometheus mode already supports multiple parachains.

PR #235, opened 2023-03-01
Track multiple parachains in the commander

vstakhov:
This addresses #193

### DIFF
--- a/src/pc/mod.rs
+++ b/src/pc/mod.rs
@@ -46,7 +46,7 @@ pub(crate) mod prometheus;
 pub(crate) mod stats;
 pub(crate) mod tracker;
 
-use crate::core::collector::CollectorUpdateEvent;
+use crate::core::collector::{CollectorStorageApi, CollectorUpdateEvent};
 use prometheus::{Metrics, ParachainCommanderPrometheusOptions};
 use tracker::ParachainBlockTracker;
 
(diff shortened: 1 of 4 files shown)


src/pc/mod.rs
```diff
@@ -68,7 +69,7 @@ pub(crate) struct ParachainCommanderOptions {
 	pub node: String,
 	/// Parachain id.
 	#[clap(long)]
-	para_id: u32,
+	para_id: Vec<u32>,
```
sandreim, line 72: From a UX perspective this doesn't seem easy to use. If we deploy on Versi with 70 parachains, it will be quite a long cmd line. We could make this field optional and if it is not specified we track all parachains that get a core assigned. This would enable us to also track parathreads without any other change.

I am hoping implementing this is easier after the collector and pc refactor. 

As an implementation I suggest to have a task that takes care of spawning and stopping instances of `watch_node`. The task would also need to keep track of the [core assignments](https://github.com/paritytech/polkadot/blob/dfa8c5ae321e0523f88fdc43afc6d743812ee49e/runtime/parachains/src/scheduler.rs#L123) and perform some pruning for parachains that didn't get a core assigned to them for a period of time.

sandreim (approved):
Nice to see the refactoring pays off :) I have just one suggestion, which might require a different PR if you think so.

src/pc/mod.rs
vstakhov: We can probably add a special parahain id `all` to subscribe to broadcast updates from the collector. However, that will require some major changes in the logic of the tracker and statistics to include per-parachain filter for everything.

vstakhov: However, this task is really a separate one: with this change we can easily take some parachains and observe them in the real-time simultaneously comparing them block-by-block. A take-them-all mode is definitely useful but it should probably use a different subscription mode.

Issue #203, opened 2023-01-11
Parachain commander: skipped slots details in summary


sandreim:
For debugging purposes, we should add an option to print out all the relay parents and their block numbers for all skipped slots.

PR #239, opened 2023-03-02
Add relay block numbers to skipped slots stats

AndreiEres:
Fixed https://github.com/paritytech/polkadot-introspector/issues/203

### DIFF
--- a/src/pc/mod.rs
+++ b/src/pc/mod.rs
@@ -76,6 +76,8 @@ pub(crate) struct ParachainCommanderOptions {
 	/// Defines subscription mode
 	#[clap(short = 's', long = "subscribe-mode", default_value_t, value_enum)]
 	pub subscribe_mode: SubxtSubscriptionMode,
+	#[clap(long = "last-skipped-slots", default_value = "10")]
+	pub last_skipped_slots: usize,
 	#[clap(flatten)]
 	collector_opts: CollectorOptions,
 	/// Mode of running - CLI/Prometheus. Default or no subcommand means `CLI` mode.
@@ -152,7 +154,13 @@ impl ParachainCommander {
 	) {
 		// The subxt API request executor.
 		let executor = api_service.subxt();
-		let mut tracker = tracker::SubxtTracker::new(para_id, self.node.as_str(), executor, api_service);
+		let mut tracker = tracker::SubxtTracker::new(
+			para_id,
+			self.node.as_str(),
+			executor,
+			api_service,
+			self.opts.last_skipped_slots,
+		);
 
 		loop {
 			match from_collector.try_recv() {
--- a/src/pc/mod.rs
+++ b/src/pc/mod.rs
@@ -76,8 +76,9 @@ pub(crate) struct ParachainCommanderOptions {
 	/// Defines subscription mode
 	#[clap(short = 's', long = "subscribe-mode", default_value_t, value_enum)]
 	pub subscribe_mode: SubxtSubscriptionMode,
-	#[clap(long = "last-skipped-slots", default_value = "10")]
-	pub last_skipped_slots: usize,
+	/// The number of last blocks with missing slots to display
+	#[clap(long = "last-skipped-slot-blocks", default_value = "10")]
+	pub last_skipped_slot_blocks: usize,
 	#[clap(flatten)]
 	collector_opts: CollectorOptions,
 	/// Mode of running - CLI/Prometheus. Default or no subcommand means `CLI` mode.
@@ -159,7 +160,7 @@ impl ParachainCommander {
 			self.node.as_str(),
 			executor,
 			api_service,
-			self.opts.last_skipped_slots,
+			self.opts.last_skipped_slot_blocks,
 		);
 
 		loop {
--- a/src/pc/stats.rs
+++ b/src/pc/stats.rs
@@ -218,10 +218,7 @@ impl ParachainStats {
 		let index = self.skipped_slots as usize % self.last_blocks_with_skipped_slots.capacity();
 		blocks.rotate_left(index);
 
-		blocks
-			.iter()
-			.map(|h| format!("\n   {} {:?}", h.0, h.1).bright_purple().to_string())
-			.collect()
+		blocks.iter().map(|h| format!("\n   {} {:?}", h.0, h.1)).collect()
 	}
 }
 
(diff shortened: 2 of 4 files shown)


src/pc/stats.rs
```diff
@@ -118,7 +118,7 @@ pub struct ParachainStats {
 	backed_count: u32,
 	/// Number of skipped slots, where no candidate was backed and availability core
 	/// was free.
-	skipped_slots: u32,
+	skipped_slots: Vec<u32>,
```
sandreim, line 121: These have to be the block hashes.

src/pc/stats.rs
```diff
@@ -194,8 +194,8 @@ impl ParachainStats {
 	}
 
 	/// Update skipped slots count
-	pub fn on_skipped_slot(&mut self) {
-		self.skipped_slots += 1;
+	pub fn on_skipped_slot(&mut self, block_number: u32) {
+		self.skipped_slots.push(block_number);
```
sandreim, line 198: If the tool runs for long, this will fill up memory. I think we should have a hard limit on how many of this hashes we keep. Ideally we should drop the oldest ones when we hit the limit to make room for newer hashes.

src/pc/stats.rs
```diff
@@ -216,8 +216,13 @@ impl Display for ParachainStats {
 		)?;
 		writeln!(
 			f,
-			"Skipped slots: {}, slow availability: {}, slow bitfields propagation: {}",
-			self.skipped_slots.to_string().bright_purple(),
+			"Skipped slots: {}, {:?}",
+			self.skipped_slots.len().to_string().bright_purple(),
```
sandreim, line 220: I prefer to have a separate line for the hashes and keep this as is.

src/pc/stats.rs
vstakhov: Yes, I was about to suggest using of the `[VecDeque](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#)` instead of the `Vec` here.

src/pc/stats.rs
vstakhov: I think that block numbers are somehow more readable. Like we have missed a slot on block `6`, `8` and `9` instead of some long hash. But I don't have a strong opinion on that.

AndreiEres: We can collect them together: `(number, hash)`.

src/pc/stats.rs
AndreiEres: @sandreim Should the amount be really small, like 10, because we use CLI.

vstakhov: It can also be started in Prometheus mode, so if a relay chain stalls, this will have hashes for all relay blocks multiplied by number of parachains.

src/pc/stats.rs
sandreim: Yeah, maybe tuple is best. I usually want the hash so I can use it to filter logs. With numbers I'd still have to translate to hash, and if there are forks it will be more work.

src/pc/stats.rs
sandreim: > @sandreim Should the amount be really small, like 10, because we use CLI.

yeah, that makes sense as a default value, but let's add a CLI option to override it.

src/pc/stats.rs
AndreiEres: Fixed it to tuples

src/pc/stats.rs
AndreiEres: Added

src/pc/stats.rs
AndreiEres: Done

src/pc/stats.rs
```diff
@@ -194,8 +197,31 @@ impl ParachainStats {
 	}
 
 	/// Update skipped slots count
```
sandreim, line 199: This comment needs to be updated.

src/pc/stats.rs
```diff
@@ -194,8 +197,31 @@ impl ParachainStats {
...
 		self.skipped_slots += 1;
+
+		let block = (update.block_number, update.block_hash);
+		if self.last_blocks_with_skipped_slots.len() < self.last_blocks_with_skipped_slots.capacity() {
+			self.last_blocks_with_skipped_slots.push(block)
+		} else {
+			let index = self.skipped_slots as usize % self.last_blocks_with_skipped_slots.capacity() - 1;
+			self.last_blocks_with_skipped_slots[index] = block;
+		}
+	}
```
sandreim, line 210: Why not use [VecDeque](https://doc.rust-lang.org/std/collections/struct.VecDeque.html) ? For each slot you just need to `push_back()` and `pop_front()` only when over max capacity.

sandreim: Would be nice if we also made a struct like `SkippedSlot { pub block_number: u32, pub block_hash: H256, }` instead of the tuple.

src/pc/mod.rs
```diff
@@ -76,6 +76,8 @@ pub(crate) struct ParachainCommanderOptions {
 	/// Defines subscription mode
 	#[clap(short = 's', long = "subscribe-mode", default_value_t, value_enum)]
 	pub subscribe_mode: SubxtSubscriptionMode,
+	#[clap(long = "last-skipped-slots", default_value = "10")]
```
sandreim, line 79: Needs doc comment to describe what the parameter does. This will show up in `--help`.

src/pc/stats.rs
```diff
@@ -194,8 +197,31 @@ impl ParachainStats {
...
+
+	fn last_blocks_string(&self) -> String {
+		if self.last_blocks_with_skipped_slots.is_empty() {
+			return String::from("-")
```
sandreim, line 214: Might be a better:
```suggestion
			return String::from("none")
```

src/pc/stats.rs
```diff
@@ -194,8 +197,31 @@ impl ParachainStats {
...
+		}
+
+		let mut blocks = self.last_blocks_with_skipped_slots.clone();
+		let index = self.skipped_slots as usize % self.last_blocks_with_skipped_slots.capacity();
+		blocks.rotate_left(index);
```
sandreim, line 219: Using `VecDeque` would allow us to remove this code.

src/pc/stats.rs
AndreiEres: Replaced with structs :-)

src/pc/stats.rs
```diff
@@ -109,6 +111,32 @@ struct DisputesStats {
...
+	block_number: u32,
+	block_hash: H256,
+}
+
+impl SkippedSlotBlock {
+	pub fn new(update: &ParachainProgressUpdate) -> Self {
+		Self { block_number: update.block_number, block_hash: update.block_hash }
+	}
+
+	pub fn join_to_string(blocks: &VecDeque<Self>) -> String {
```
sandreim, line 125: This should be just a function, doesn't need to be part of `SkippedSlotBlock` impl.

sandreim (approved):

vstakhov (approved):

Issue #201, opened 2023-01-09
Parachain commander: show current parachain host configuration


sandreim:
When starting up it would be nice to show some of the parachain host configuration fields which impact performance, such as:
- needed_approvals
- max_validators_per_core
- max_validators
- no_show_slots
- n_delay_tranches

sandreim:
implemented in https://github.com/paritytech/polkadot-introspector/pull/242

PR #242, opened 2023-03-03
PC: show host configuration on startup

vstakhov:
This PR allows to show a host configuration on startup. Unfortunately, this call is very specific to each network, so it might probably a good idea to postpone it unless we have a way to make an `ok-to-fail` subxt calls.

Issue: #201


sandreim (approved):

sandreim:
Nice! Thanks, I really needed this !

sandreim:
A metadata update is also included here, is this intentional ?

vstakhov:
Yes, this exact method is very metadata dependent :( E.g. metadata from Versi can work on Kusama and Polkadot. However, the HostConfig is very different for each network, meaning that even Polkadot metadata cannot work in Kusama and vice-versa.

PR #255, opened 2023-03-09
Move block events mapping to collector

AndreiEres:


src/core/collector/mod.rs
```diff
@@ -203,6 +212,24 @@ impl Collector {
 		})
 	}
 
+	pub async fn map_subxt_event(&mut self, event: &SubxtEvent) -> color_eyre::Result<Vec<SubxtEvent>> {
```
vstakhov, line 215: Neat: `map` usually means `Iter`-like semantics, so probably it is better to call it `transform_subxt_event`. Do we really need `self` here, especially as `mut`?

src/core/collector/mod.rs
```diff
@@ -203,6 +212,24 @@ impl Collector {
...
+		match event {
+			SubxtEvent::NewHead(block_hash) => {
+				let block_hash = *block_hash;
+				let mut subxt_events = vec![SubxtEvent::NewHead(block_hash)];
+				let client = OnlineClient::<PolkadotConfig>::from_url(self.endpoint.as_str()).await?;
```
vstakhov, line 220: I'm not quite sure that it is a good idea to make this call directly. We have the executor service in the collector that can use the connections pool and it can wrap all retries logic/backpressure et all. I think it is better to use it instead of opening an additional connection to an RPC node. Furthermore, `subxt` subscription provides these events for *free*, so why do we need to query them explicitly in the first place?

src/core/collector/mod.rs
sandreim, line 215: The naming is confusing a bit wrt return type. A doc comment describing how it works would be helpful.

Looking at the code, it seems to parse a `SubxtEvent` and produce more `SubxtEvents`. This sounds a bit odd. :) I think we need to split them up, such that block events are no longer mixed with `subxt` events and this function actually decodes all block events, which makes sense.

src/core/collector/mod.rs
sandreim: I think we should move away from using the API service via channel. We need to get rid of the codebloat that comes with it. I am thinking that the API service as a global would be really neat :P

vstakhov: It is not done via channels now. But there is still a logic around reconnections and backpressure that should be reused. Nevertheless, I still think that we should use events from subxt subscription and not ask a node for them one more time.

src/core/collector/mod.rs
AndreiEres: I need the endpoint over self. `mut` is useless, true

AndreiEres: I agree. I wanted to change it but hesitated to use an event with a single value.

vstakhov (approved):
LGTM!

src/core/collector/mod.rs
```diff
@@ -203,13 +212,32 @@ impl Collector {
 		})
 	}
 
-	/// Process a next subxt event
-	pub async fn process_subxt_event(&mut self, event: &SubxtEvent) -> color_eyre::Result<()> {
+	/// Collects chain events from new head including block events parsing
+	pub async fn collect_chain_events(&mut self, event: &SubxtEvent) -> color_eyre::Result<Vec<ChainEvent>> {
```
sandreim, line 216: 👍

sandreim (approved):
Nice work @AndreiEres !

Issue #197, opened 2023-01-06
Parachain commander: publish finality lag


sandreim:
In https://github.com/paritytech/polkadot-introspector/pull/181 we added support for selecting to which block stream we subscribe. In order to compute finality lag we should follow both finalized and best streams but the processing of blocks should be done as configured in the CLI

AndreiEres:
### Some notes
```
relay chain finality lag = best block number - last finalized block number
```
this functionality could best live in the collector, parachain commander would just display it.

we can compute finality lag at each best block and put that into storage 

`async fn process_new_head(&mut self, block_hash: H256) -> color_eyre::Result<()>` — puts block header in storage

we could change the storage entry to also contain the finality lag or it is far better to have just the last finalized block

from Parachain commander we just read the storage and get that info
```

AndreiEres:
@sandreim @vstakhov 
Because the implementation causes tons of inner changes, I'd rather split it for few pull requests. The main idea is to move block and events handling to the collector. Then we can remove the best/finalized mode from subxt subscription and start receiving both block types and forward them to the collector. In the collector, we can extract all the needed data from blocks and produce events according to the best/finalized mode.

### Steps
- [ ] Move block events mapping to collector
- [ ] Start to receive both block types from subxt and handle it in the collector
- [ ] Add the lag calculation

sandreim:
> @sandreim @vstakhov Because the implementation causes tons of inner changes, I'd rather split it for few pull requests. 

This makes sense to me!

>The main idea is to move block and events handling to the collector. Then we can remove the best/finalized mode from subxt subscription and start receiving both block types and forward them to the collector. In the collector, we can extract all the needed data from blocks and produce events according to the best/finalized mode.
> 

This is similar to what I had in mind.

PR #259, opened 2023-03-13
Move subscribe-mode to collector

AndreiEres:
Part of https://github.com/paritytech/polkadot-introspector/issues/197
Because we need to handle both types of blocks I changed subscription following the example https://github.com/paritytech/subxt/blob/lexnv/diff_best_last_finalized/examples/examples/subscribe_chain_head_events.rs#L58-L84

### DIFF
--- a/src/core/subxt_subscription.rs
+++ b/src/core/subxt_subscription.rs
@@ -111,7 +111,7 @@ impl SubxtSubscription {
 							FollowEvent::Finalized(finalized) => {
 								for hash in finalized.finalized_block_hashes.iter() {
 									info!("[{}] Finalized block imported ({:?})", url, hash);
-									if let Err(e) = update_channel.send(SubxtEvent::NewBestHead(*hash)).await {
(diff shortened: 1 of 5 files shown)


src/core/subxt_subscription.rs
```diff
@@ -17,41 +17,34 @@
...
-	/// New relay chain head
-	NewHead(<PolkadotConfig as subxt::Config>::Hash),
+	/// New relay chain best head
+	NewBestHead(<PolkadotConfig as subxt::Config>::Hash),
+	/// New relay chain finalized head
+	NewFinalizedHead(<PolkadotConfig as subxt::Config>::Hash),
 }
 
-pub struct SubxtWrapper {
+pub struct SubxtSubscription {
```
vstakhov, line 40: 👍

vstakhov (approved):
It looks very good to me, since we can now follow multiple chains by subxt events.

Issue #208, opened 2023-01-19
Add tests to detect when metadata needs to be updated


sandreim:
Currently we discover this when using the tool for development. We should be proactive about it, but initially having some tests would help spotting the issue on PRs.

vstakhov:
I have just found the same issue when working on #201
We also need to deal with subxt errors in a more graceful way, as some of them are apparently not fatal.

PR #274, opened 2023-03-20
Add metadata checker

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/208

- [x] Add MetadataChecker
- [x] Add CI rules to check metadata (CI should fail)
- [x] Update metadata (CI should succeed)


sandreim (commented):
I think this approach is good. Indeed this needs to be run per runtime we support (via feature).

vstakhov (approved):

alvicsam (approved):

PR #278, opened 2023-03-20
Add versi as a feature to CI pipeline

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/276

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -83,6 +83,7 @@ clippy:
   script:
     - time cargo +nightly clippy --no-default-features --all-targets --features=polkadot -- -D warnings
     - time cargo +nightly clippy --no-default-features --all-targets --features=rococo -- -D warnings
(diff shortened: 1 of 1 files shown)


sandreim:
I think we should remove the `versi` feature, it's same as `rococo`. Versi testnet uses the `rococo` runtime but with different parameters, for example the duration of a session is just 10m on Versi vs 1h on Rococo

sandreim:
There was a discussion about it having a separate runtime at some point and maybe that's why we have it in the code ...

AndreiEres:
> I think we should remove the `versi` feature, it's same as `rococo`. Versi testnet uses the `rococo` runtime but with different parameters, for example the duration of a session is just 10m on Versi vs 1h on Rococo

So we need to set `rococo` as default?

AndreiEres:
@sandreim cut `versi`

sandreim (approved):

Issue #283, opened 2023-03-23
Telemetry support


AndreiEres:
We'd like to add some additional off-chain information from [the telemetry](https://telemetry.polkadot.io/). The telemetry itself is an event stream. There is no state, so we need to subscribe to it and save the current state in our storage. We can start with collecting existing events, but then we can also add support of additional events related to parachains work.

### Progress
The work is expected to be large and can be tracked in the following tasks:
- [ ] https://github.com/paritytech/polkadot-introspector/issues/178
- [ ] https://github.com/paritytech/polkadot-introspector/issues/98
- [ ] https://github.com/paritytech/polkadot-introspector/issues/91

sandreim:
https://github.com/paritytech/polkadot-introspector/issues/91 is out of date, I'll update the ticket to current vision

PR #284, opened 2023-03-24
Add simple telemetry feed

AndreiEres:
Related to https://github.com/paritytech/polkadot-introspector/issues/283
Adds a subscription to telemetry feed. Could be used as a scaffolding for further work.


src/core/telemetry_feed.rs
```diff
@@ -51,8 +68,16 @@ impl TelemetryFeed {
 			},
 			// TODO: Add the following messages
 			//  3: AddedNode
-			//  4: RemovedNode
-			//  5: LocatedNode
+			// RemovedNode
+			4 => {
+				let node_id = serde_json::from_str(raw_payload.get())?;
+				TelemetryFeed::RemovedNode { node_id }
```
sandreim, line 74: Do we need these kind of events ?

AndreiEres:
Here are all messages defined there https://github.com/paritytech/substrate-telemetry/blob/b25f3fbf2270c009aafebefa10bdcc034b8d1cc0/backend/telemetry_core/src/feed_message.rs#L105-L125

Then we can check which telemetry events are not being processed and add additional support.

introspector/src/main.rs
```diff
@@ -137,7 +144,38 @@ async fn main() -> color_eyre::Result<()> {
...
+	shutdown_tx: Sender<()>,
+) -> Vec<tokio::task::JoinHandle<()>> {
+	futures.push(tokio::spawn(on_shutdown(shutdown_tx)));
+	futures
+}
+
+async fn on_shutdown(shutdown_tx: Sender<()>) {
+	signal::ctrl_c().await.unwrap();
+	let _ = shutdown_tx.send(());
+}
```
vstakhov, line 176: 👍

introspector/src/core/telemetry_subscription.rs
```diff
@@ -0,0 +1,113 @@
...
+use futures_util::StreamExt;
+use log::info;
+use tokio::{
+	net::TcpStream,
+	sync::{
+		broadcast::Sender as BroadcastSender,
+		mpsc::{error::SendError, Sender},
+	},
+};
+use tokio_tungstenite::{connect_async, tungstenite::Message, MaybeTlsStream, WebSocketStream};
```
vstakhov, line 29: Why don't we use `jsonrpsee` as `subxt` does?

vstakhov: It seems we cannot, and the `tokio_tungstenite` is a de-facto standard when you just need to subscribe on events through a websocket.

vstakhov (approved):

introspector/src/core/telemetry_subscription.rs
```diff
@@ -0,0 +1,119 @@
...
+						_ => continue,
+					};
+					let feed = TelemetryFeed::from_bytes(&bytes);
+					if let Err(e) = feed {
+						on_error(e);
+						continue;
+					}
+
+					for message in feed.unwrap() {
+						info!("[telemetry] {:?}", message);
```
sandreim, line 70: I think this will spam a lot should be debug.

introspector/src/core/telemetry_subscription.rs
```diff
@@ -0,0 +1,119 @@
...
+	/// One sender per consumer per URL.
+	consumers: Vec<Sender<TelemetryEvent>>,
+}
+
+impl TelemetrySubscription {
+	pub fn new(consumers: Vec<Sender<TelemetryEvent>>) -> Self {
+		Self { consumers }
+	}
+
+	// Sets up per websocket tasks to handle updates and reconnects on errors.
```
sandreim, line 45: This comment seems to not reflect what the function does. This is actually the task.

introspector/src/telemetry/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Clone, Debug, Parser)]
+#[clap(rename_all = "kebab-case")]
+pub(crate) struct TelemetryOptions {
+	/// Web-Socket URL of a telemetry backend
+	#[clap(name = "ws", long)]
+	pub url: String,
+	// Chain's genesis hash
+	#[clap(name = "chain", long)]
+	pub chain_hash: H256,
```
sandreim, line 34: Why do we need this ?

introspector/src/telemetry/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+use clap::Parser;
+use priority_channel::{Receiver, TryRecvError};
+use subxt::utils::H256;
+use tokio::sync::broadcast::Sender as BroadcastSender;
+
+use crate::core::{TelemetryEvent, TelemetrySubscription, MAX_MSG_QUEUE_SIZE};
+
+#[derive(Clone, Debug, Parser)]
+#[clap(rename_all = "kebab-case")]
+pub(crate) struct TelemetryOptions {
```
sandreim, line 28: So this is actually a very simple CLI telemetry frontend. I'd add at least some filter by node id, otherwise the output would be too much for a large net.

introspector/src/telemetry/mod.rs
AndreiEres: Telemetry in the beginning sends events only about added chains. We need to subscribe to one of them to start getting its events.

introspector/src/core/telemetry_subscription.rs
AndreiEres: It's true

introspector/src/telemetry/mod.rs
sandreim: Ah, this looks like usability trouble. In practice do we have multiple chains on the same backend ?

introspector/src/telemetry/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+		futures.push(tokio::spawn(Self::watch(self.update_channel, self.opts.network_id)));
+
+		Ok(futures)
+	}
+
+	async fn watch(update: Receiver<TelemetryEvent>, network_id: String) {
+		let mut node_id: Option<FeedNodeId> = None;
+
+		loop {
+			match update.try_recv() {
```
sandreim, line 77: Why not just do `update.recv().await` and block until a message arrives or an error happens.

sandreim (approved):

PR #295, opened 2023-03-30
Move storage to common package

AndreiEres:


sandreim (commented):
What is the reasoning for this change ? I would have expected this crate  to be named `storage` or just move out the `core` as a crate. The storage is a core functionality.

AndreiEres:
> What is the reasoning for this change ? I would have expected this crate to be named `storage` or just move out the `core` as a crate. The storage is a core functionality.

Sorry I wasn't explicit. I am going to move all core as a package but I didn't want to do it in one time to prevent some hidden bugs. So I started with just a storage.

essentials/src/storage.rs
```diff
@@ -164,6 +164,10 @@ pub trait RecordsStorage<K> {
 		K: Borrow<Q>;
 	/// Size of the storage
 	fn len(&self) -> usize;
+	/// Checks if the storage is empty
+	fn is_empty(&self) -> bool {
```
sandreim, line 168: This change seems to not be related to the PR.

sandreim (approved):
LGTM

PR #300, opened 2023-04-11
Move metadata-checker to a package

AndreiEres:


alvicsam (approved):

sandreim (approved):

vstakhov:
I think we should have a single binary that is built from multiple crates (so we can hide unnecessary subcommands with feature flags if needed). With this approach we will end up with multiple binaries. I think the approach with a single binary is easier to maintain. What's your opinion here @sandreim ?

sandreim:
I am thinking to move away from a swiss army knife approach towards individual tools build on top of an introspection framework. I also believe it would be easier to maintain individual tools and provide a simpler better UX for using them.

vstakhov:
I have only one objection: when you have a single tool, it is easier to get all subcommands, by something like `polkadot-introspector --help`. In case of multiple tools it is much more complicated. Probably, we should just have some common prefix for all tools, like `polkadot-introspector-pc`, `polkadot-introspector-kvdb` etc. This makes it easier to use completion but adds quite a lot of verbosity. However, it is better than having a set of unrelated tools from my point of view.

PR #311, opened 2023-04-14
Move kvdb tool to a separate binary

vstakhov:
We also want devops to adjust their containers to use kvdb as a side command, and not a part of `polkadot-introspector`. The prefix `polkadot-` in `polkadot-kvdb` is a topic for discussions.


vstakhov:
> I think README also needs to be updated

It's done now

kvdb/src/main.rs
```diff
@@ -168,7 +165,7 @@ pub(crate) enum KvdbDumpMode {
 }
 
 #[derive(Clone, Debug, Parser)]
-#[clap(rename_all = "kebab-case")]
+#[clap(author, version, about = "Examine Polkadot and Substrate key-value databases", rename_all = "kebab-case")]
```
sandreim, line 168: I think this tool would work regardless of the software that crates the DBs.
```suggestion
#[clap(author, version, about = "Examine RocksDB and ParityDB databases", rename_all = "kebab-case")]
```

sandreim (approved):

Issue #12, opened 2022-02-25
Add exponential backoff for retries


sandreim:
We also need to expose it's settings in cli

PR #312, opened 2023-04-14
Add retry config

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/12

### DIFF
--- a/essentials/src/utils.rs
+++ b/essentials/src/utils.rs
@@ -56,7 +56,6 @@ impl Default for Retry {
 impl Retry {
 	pub fn new() -> Self {
 		let opts = RetryOptions::parse();
-		println!("{:?}", opts);
 
 		Self { count: 0, max_count: opts.max_count, delay: opts.retry_delay }
 	}
--- a/essentials/src/utils.rs
+++ b/essentials/src/utils.rs
@@ -21,7 +21,7 @@ use std::time::Duration;
 use thiserror::Error;
 use tokio::time::sleep;
 
-#[derive(Clone, Debug, Parser, Default)]
+#[derive(Clone, Debug, Parser)]
 #[clap(
 	rename_all = "kebab-case",
 	allow_external_subcommands = true, // HACK: to parse it as a standalone config
@@ -45,6 +45,12 @@ pub struct Retry {
 pub enum RetryError {
 	#[error("Max count reached")]
 	MaxCountReached,
+}
+
+impl Default for Retry {
+	fn default() -> Self {
+		Self::new()
+	}
 }
 
 impl Retry {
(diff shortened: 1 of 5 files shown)


essentials/src/utils.rs
```diff
@@ -15,17 +15,30 @@
...
+	rename_all = "kebab-case",
+	allow_external_subcommands = true, // HACK: to parse it as a standalone config
+)]
+pub struct RetryOptions {
+	/// Max of times to retry a failed connection before giving up.
+	#[clap(name = "retry", default_value = "3", long)]
+	max_count: u32,
+	/// Delay in ms to wait between retry attempts
+	#[clap(default_value = "100", long)]
+	retry_delay: u32,
```
vstakhov, line 35: I usually prefer time intervals expressed in seconds with a floating point. But it might be probably ok to use milliseconds here.

essentials/src/utils.rs
```diff
@@ -34,16 +47,29 @@ pub enum RetryError {
...
+impl Default for Retry {
+	fn default() -> Self {
+		Self::new()
+	}
+}
+
 impl Retry {
+	pub fn new() -> Self {
+		let opts = RetryOptions::parse();
+		println!("{:?}", opts);
```
vstakhov, line 59: Isn't it a debug?

essentials/src/utils.rs
AndreiEres: We need to answer the topic starter :-)
@sandreim what do you think?

sandreim: I have no personal preference in this case. One downside of floating point numbers is that they are approximate and we should refrain from ever using them in this industry even for tools.

essentials/src/utils.rs
```diff
@@ -15,17 +15,30 @@
...
 
-use crate::constants::{RETRY_COUNT, RETRY_DELAY_MS};
+#[derive(Clone, Debug, Parser)]
+#[clap(
+	rename_all = "kebab-case",
+	allow_external_subcommands = true, // HACK: to parse it as a standalone config
+)]
+pub struct RetryOptions {
+	/// Max of times to retry a failed connection before giving up.
+	#[clap(name = "retry", default_value = "3", long)]
```
sandreim, line 31: Isn't this too low ? RPC nodes sometimes can be overloaded.

sandreim (approved):

essentials/src/utils.rs
AndreiEres: So let's leave ms here.

essentials/src/utils.rs
AndreiEres: Changed to 10

PR #315, opened 2023-04-19
Choose chain from the list in telemetry feed

AndreiEres:
Now we have to pass a genesis hash to subscribe to a particular chain in the telemetry feed what could be annoying. So I changed it to choose the chain from a list on start.

### DIFF
--- a/essentials/src/telemetry_subscription.rs
+++ b/essentials/src/telemetry_subscription.rs
@@ -24,7 +24,7 @@ use itertools::Itertools;
 use log::{debug, info, warn};
 use priority_channel::{SendError, Sender};
 use std::{
-	cmp::Reverse,
+	cmp::{min, Reverse},
 	collections::HashMap,
 	io::{stdin, BufRead},
 };
@@ -37,7 +37,7 @@ struct TelemetryStream {
 
 impl TelemetryStream {
 	async fn connect(url: &str) -> Self {
-		println!("Connecting to the telemetry server on {}", url);
+		println!("Connecting to the telemetry server on {}\n", url);
 		let (ws_stream, _) = connect_async(url).await.expect("Failed to connect");
 		Self { ws_stream }
 	}
@@ -145,6 +145,8 @@ impl TelemetrySubscription {
 	}
 }
 
+const CHAINS_CHUNK_SIZE: usize = 10;
+
 async fn choose_chain(chains: &HashMap<H256, AddedChain>) -> color_eyre::Result<H256, ()> {
 	let list: Vec<AddedChain> = chains
 		.iter()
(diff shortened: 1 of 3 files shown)


essentials/src/telemetry_subscription.rs
```diff
@@ -15,17 +15,50 @@
...
+				Ok(())
+			},
+			Err(e) => {
+				info!("Cannot subscribe to chain with hash {}: {:?}", chain, e);
+				Err(e.into())
+			},
+		}
+	}
+
+	fn next(&mut self) -> Next<'_, WebSocketStream<MaybeTlsStream<TcpStream>>> {
```
AndreiEres, line 58: I'm not sure it's idiomatic here

vstakhov: It should be a part of another trait, specifically [Stream](https://docs.rs/futures/latest/futures/stream/trait.Stream.html) trait.

vstakhov: So you need to implement `Stream` methods and that will allow to use `StreamExt` methods, such as `next`.

essentials/src/telemetry_subscription.rs
```diff
@@ -84,27 +133,84 @@ impl TelemetrySubscription {
...
+			println!("No chains found...");
+			return Err(())
+		},
+		1 => {
+			println!("Found 1 chain.\n{}", list[0]);
+			return Ok(list[0].genesis_hash)
+		},
+		size => {
+			println!("Found {} chains.\n", size);
+		},
```
vstakhov, line 169: See comment above for the `Display` trait. Error should also be meaningful and not a tuple, as in general, it is a caller responsibility what to do with a `Result` returned.

essentials/src/telemetry_subscription.rs
```diff
@@ -84,27 +133,84 @@ impl TelemetrySubscription {
...
+		for (i, chain) in indexed_list[cursor..min(cursor + CHAINS_CHUNK_SIZE, list_size)].iter() {
+			println!("{}. {}", i, chain);
+		}
+		println!(
+			"\nInput the number of a chain you want to follow (1-{}).\nENTER to loop throw the list, Q to exit.\n",
+			list_size
+		);
+		stdin().lock().read_line(&mut buf).expect("Failed to read line");
+
+		if buf == *"\n" {
```
vstakhov, line 186: I'd just trim the buffer and if it has 0 size assume a default choice.

essentials/src/telemetry_subscription.rs
```diff
@@ -47,15 +80,16 @@ impl TelemetrySubscription {
 	async fn run_per_consumer(
 		mut update_channel: Sender<TelemetryEvent>,
 		url: String,
-		chain_hash: H256,
 		shutdown_tx: BroadcastSender<()>,
 	) {
 		let mut shutdown_rx = shutdown_tx.subscribe();
-		let mut ws_stream = telemetry_stream(&url, chain_hash).await;
+		let mut stream = TelemetryStream::connect(&url).await;
+		let mut current_chain: H256 = H256::zero();
```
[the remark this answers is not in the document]
AndreiEres: To store a state and don't ask about chain every cycle. Can change it to a flag.

vstakhov (approved):

PR #323, opened 2023-04-24
Add polkadot-block-time and polkadot-kvdb to the docker image

AndreiEres:
As we extracted the polkadot-block-time to a package we need to store it additionally in our docker image.


vstakhov:
Maybe we should also add `polkadot-kvdb` there?

AndreiEres:
> Maybe we should also add `polkadot-kvdb` there?

I need to add the block-time because we deploy it. But it could be a good thing to add all utils for external use to the image.

alvicsam (approved):

vstakhov (approved):

Issue #313, opened 2023-04-18
Log parachain ids we're tracking


AndreiEres:
There is a comment from devops https://github.com/paritytech/devops/issues/2469#issuecomment-1491995834.

> It doesn't print which parachains are monitored. So it's not easy to see if that it's working properly in prometheus mode as it doesn't output logs.

As we can track all parachains could be great to log their ids in the beginning.

PR #331, opened 2023-04-25
Update all parachains tracing

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/313

### DIFF
--- a/introspector/src/pc/mod.rs
+++ b/introspector/src/pc/mod.rs
@@ -127,9 +127,13 @@ impl ParachainCommander {
 		}
 
 		println!(
-			"{} will trace parachain(s) {} on {}\n{}",
+			"{} will trace {} on {}\n{}",
 			"Parachain Commander(TM)".to_string().purple(),
-			self.opts.para_id.iter().join(","),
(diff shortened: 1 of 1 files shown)


introspector/src/pc/mod.rs
```diff
@@ -249,6 +253,7 @@ impl ParachainCommander {
 								let to_tracker = trackers.entry(para_id).or_insert_with(|| {
 									let (tx, rx) = channel_with_capacities(collector::COLLECTOR_NORMAL_CHANNEL_CAPACITY, 1);
 									futures.push(ParachainCommander::watch_node_for_parachain(self.clone(), rx, para_id, api_service.clone()));
+									info!("Added tracker for parachain {}", para_id);
```
vstakhov, line 256: We might probably add the similar log line if a parachain has been evicted (if it is not there yet of course).

vstakhov (approved):

introspector/src/pc/mod.rs
AndreiEres: It already has a message on evicting.

PR #342, opened 2023-05-02
Add search by session and validator indices to whois

AndreiEres:


whois/README.md
```diff
@@ -1,9 +1,13 @@
 # polkadot-whois
 
-Tracking of validators using a substrate telemetry data.
+Tracking of validators using a substrate telemetry data. The data can be queried by validator's `AccountId` or `(session_index, validator_index)`.
```
sandreim, line 3: Now also uses on-chain data, we should reflect that here.

PR #350, opened 2023-05-05
Exit if cannot process tracker update

AndreiEres:


essentials/src/utils.rs
```diff
@@ -61,6 +61,7 @@ impl Retry {
 	pub async fn sleep(&mut self) -> color_eyre::Result<(), RetryError> {
 		self.count += 1;
 		if self.count > self.max_count {
+			println!("Here is where we're going to raise");
```
sandreim, line 64: Is this intentional?

sandreim (approved):

essentials/src/utils.rs
AndreiEres: No, sorry, forgot to remove.

PR #356, opened 2023-05-09
Fix warning about borrowing while type conververtion

AndreiEres:


sandreim:
Wondering if we still have a need for Jaeger at all here 😁

vstakhov (approved):
Yes, cloning is a simple and not very expensive solution for this particular case.

Issue #368, opened 2023-05-19
Improve accuracy of tracer block time metrics


sandreim:
While analyzing the data collected via https://grafana.parity-mgmt.parity.io/goto/Vz-melU4R?orgId=1 it seems that we might have inaccurate metrics when rpc connection goes south as well due to scraping intervals, especially for histograms. 

We should explore some ways of improving on that.

sandreim:
Maybe it is worth to have separate metrics for para block times in seconds also. This would make it more accurate as now we do multiply those by the relay block time.

AndreiEres:
We added para block times in seconds recently. @sandreim What else do you think we could have done?

sandreim:
I'd compare the numbers we get for kusama statemine block times (via collator metrics) with numbers we get from introspector for statemine.

Query for statemine metrics: `1/min(rate(substrate_block_height{chain="statemine"}[5m]) > 0) by (chain)`

If these are in sync we can close the issue otherwise I'd say we must dive deeper

AndreiEres:
<img width="1462" alt="image" src="https://github.com/paritytech/polkadot-introspector/assets/27277055/f4be6e10-3de6-491c-8c85-7c0ca348767b">
https://grafana.parity-mgmt.parity.io/goto/6CT_D3uVg?orgId=1

On average, we have ~0.5s difference between metrics. If we consider collator's metrics as realistic, we should keep improving the accuracy.

PR #370, opened 2023-05-22
Add candidate block time in seconds

AndreiEres:
Part of https://github.com/paritytech/polkadot-introspector/issues/368

### DIFF
--- a/parachain-tracer/src/prometheus.rs
+++ b/parachain-tracer/src/prometheus.rs
@@ -67,7 +67,7 @@ struct MetricsInner {
 	/// Average candidate inclusion time measured in relay chain blocks.
 	para_block_times: HistogramVec,
 	/// Average candidate inclusion time measured in seconds.
-	para_block_time_sec: HistogramVec,
+	para_block_times_sec: HistogramVec,
 	/// Finality lag
 	finality_lag: Gauge,
 }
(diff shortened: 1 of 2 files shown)


parachain-tracer/src/prometheus.rs
```diff
@@ -66,6 +66,8 @@ struct MetricsInner {
 	bitfields: IntGaugeVec,
 	/// Average candidate inclusion time measured in relay chain blocks.
 	para_block_times: HistogramVec,
+	/// Average candidate inclusion time measured in seconds.
+	para_block_times_sec: HistogramVec,
```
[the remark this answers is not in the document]
vstakhov: I think we don't, since this value is a time serie in fact, so each time there can be merely one measurement.

parachain-tracer/src/tracker.rs
```diff
@@ -710,6 +720,13 @@ impl SubxtTracker {
 		std::time::Duration::from_millis(cur_ts).saturating_sub(std::time::Duration::from_millis(base_ts))
 	}
 
+	/// Returns the time for the current candidate
+	pub fn get_candidate_time(&self) -> std::time::Duration {
+		let current = self.current_relay_block_ts.unwrap_or_default();
+		let base = self.last_included_at_ts.unwrap_or(current);
```
vstakhov, line 726: This will lead to the block time = 0 when there is no previous block. We should either not update the counter in this case or set it to undefined (e.g. `NaN` but I'm not sure how nan is treated by prometheus).

AndreiEres: We don't write the metric if there is no previous block. Same as for `para_block_time`.

sandreim: This function should return an Option to prevent metric updates when we don't have last block.

PR #373, opened 2023-05-25
Add kusama feature

AndreiEres:
Kusama updated its metadata and now it's diferrent from the polkadot's one. So that we need to add support for both of them at least until they use same metadata again.


alvicsam (approved):

sandreim:
Best practice is add a descriptive comment for the reasoning of the PR, instead of leaving that field blank. In this case I know why we need this, since we discussed it offline, but others don't

sandreim (approved):

.gitlab-ci.yml
```diff
@@ -126,12 +136,15 @@ build:
   <<:                              *collect-artifacts
   script:
     - mkdir ./artifacts || echo "dir exists"
-    - mkdir ./artifacts/polkadot || echo "dir exists"
     - mkdir ./artifacts/rococo || echo "dir exists"
-    - time cargo build --release --no-default-features --features=polkadot
-    - cp target/release/{polkadot-block-time,polkadot-jaeger,polkadot-kvdb,polkadot-parachain-tracer,polkadot-whois} ./artifacts/polkadot/
+    - mkdir ./artifacts/kusama || echo "dir exists"
```
alvicsam, line 140: consider using `mkdir -p`

PR #374, opened 2023-05-25
Bump v0.2.7

AndreiEres:


alexggh (approved):
Why is the bump needed for this case ?

AndreiEres:
> Why is the bump needed for this case ?

We deploy on new tags. So we bump version and then add new tag to trigger the deployment. Now we need to fix the issue with metadata.

Issue #92, opened 2022-08-24
Replace `String` usage with `&str` 


sandreim:
There are more than one occurrences such as this:
https://github.com/paritytech/polkadot-introspector/blob/master/src/core/api/subxt_wrapper.rs#L72

PR #390, opened 2023-06-12
Stop using strings in function arguments

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/92

### DIFF
--- a/essentials/src/chain_head_subscription.rs
+++ b/essentials/src/chain_head_subscription.rs
@@ -202,7 +202,7 @@ impl ChainHeadSubscription {
 			.into_iter()
 			.zip(urls.into_iter())
 			.map(|(update_channel, url)| {
-				tokio::spawn(Self::run_per_node(update_channel, url.to_string(), shutdown_tx.clone(), retry.clone()))
+				tokio::spawn(Self::run_per_node(update_channel, url, shutdown_tx.clone(), retry.clone()))
 			})
 			.collect()
 	}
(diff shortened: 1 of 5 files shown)


AndreiEres:
> Does this cover all usages ?

I removed all usages of `String` in function args. 
There are still strings in `Struct` fields and in return types of functions, but I don't thing it's relevant to the issue. What do you think?

essentials/src/telemetry_subscription.rs
```diff
@@ -80,6 +80,8 @@ impl TelemetrySubscription {
 	// Subscribes to a telemetry feed handling graceful shutdown.
 	async fn run_per_consumer(
 		mut update_channel: Sender<TelemetryEvent>,
+		// Usually, we avoid using String in function arguments to prevent unnecessary cloning.
```
AndreiEres, line 83: @sandreim check it out

essentials/src/telemetry_subscription.rs
```diff
@@ -80,6 +80,8 @@ impl TelemetrySubscription {
...
+		// However, we spawn this particular method as an asynchronous task, so we make an exception here.
 		url: String,
```
AndreiEres, line 85: @sandreim check it out

essentials/src/chain_head_subscription.rs
```diff
@@ -103,7 +103,7 @@ impl ChainHeadSubscription {
 	// Per consumer
 	async fn run_per_node(
 		mut update_channel: Sender<ChainHeadEvent>,
-		url: String,
+		url: String, // `String` rather than `&str` because we spawn this method as an asynchronous task
```
AndreiEres, line 106: @sandreim but we still have to use strings for methods we run as a task, like that

Issue #115, opened 2022-09-20
Historical mode - run on older blocks rather than following the tip of chain


sandreim:
For example we could answer questions like - at which block was a candidate included/disputed... etc.

PR #393, opened 2023-06-15
Add historical mode to parachain-tracer

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/115

To achieve that we mimic chain subscription sending block hashes from appropriate blocks to collector.

### DIFF
--- a/parachain-tracer/README.md
+++ b/parachain-tracer/README.md
@@ -37,7 +37,9 @@ It's important to note that the tool cannot further drill down into the details 
 
 Assuming you have cloned the repo, and built the binary, all you need for it to work is a relay chain RPC endpoint.
 
-Example: `polkadot-parachain-tracer --ws=wss://kusama-try-runtime-node.parity-chains.parity.io:443 --para-id=2107`
+Example: `polkadot-parachain-tracer --ws wss://rpc.polkadot.io:443 --para-id 2107 cli`
+
+It is possible to run the tool in historical mode to trace parachains between specific blocks instead of following live chain progress: `polkadot-parachain-tracer --ws wss://rpc.polkadot.io:443 --para-id 2107 --historical --from 16080000 --to 16080050 cli`
 
 ```
 USAGE:


essentials/src/historical_subscription.rs
```diff
@@ -0,0 +1,183 @@
...
+		urls: Vec<String>,
+		from_block_number: BlockNumber,
+		to_block_number: BlockNumber,
+		retry: RetryOptions,
+	) -> HistoricalSubscription {
+		HistoricalSubscription { urls, from_block_number, to_block_number, consumers: Vec::new(), retry }
+	}
+
+	// Per consumer
+	async fn run_per_node(
```
sandreim, line 109: Mismatch with comment above

essentials/src/historical_subscription.rs
```diff
@@ -0,0 +1,183 @@
...
+						return;
+					}
+				}
+			}
+
+			let message = executor.get_block_hash(&url, Some(block_number)).await;
+			let block_hash = match message {
+				Ok(Some(v)) => v,
+				Ok(None) => {
+					error!("Subscription to {} failed, block hash for block #{} not found", url, block_number);
```
sandreim, line 142: How hard are we retrying here ?

essentials/src/historical_subscription.rs
```diff
@@ -0,0 +1,183 @@
...
+		to_block_number: BlockNumber,
+		shutdown_tx: BroadcastSender<()>,
+		retry: RetryOptions,
+	) {
+		let mut shutdown_rx = shutdown_tx.subscribe();
+		let mut executor = RequestExecutor::new(retry);
+		const HEARTBEAT_INTERVAL: Duration = Duration::from_millis(200);
+		let mut heartbeat_periodic = interval_at(tokio::time::Instant::now() + HEARTBEAT_INTERVAL, HEARTBEAT_INTERVAL);
+
+		for block_number in from_block_number..=to_block_number {
```
sandreim, line 122: How are we handling `to_block_number` being actually greater than the tip of the chain ?

essentials/src/chain_subscription.rs
```diff
@@ -0,0 +1,28 @@
...
+//
+
+use subxt::PolkadotConfig;
+
+#[derive(Debug)]
+pub enum ChainSubscriptionEvent {
+	/// New relay chain best head
+	NewBestHead(<PolkadotConfig as subxt::Config>::Hash),
+	/// New relay chain finalized head
+	NewFinalizedHead(<PolkadotConfig as subxt::Config>::Hash),
```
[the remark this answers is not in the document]
AndreiEres: Good catch, I'll rename it.

essentials/src/chain_subscription.rs
```diff
@@ -0,0 +1,28 @@
...
+	/// Heartbeat event
+	Heartbeat,
```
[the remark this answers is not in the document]
AndreiEres: Actually it's not relevant to the PR, I just moved types out to use it in different subscription types.

essentials/src/historical_subscription.rs
AndreiEres: Every subxt call wrapped into retry.

essentials/src/historical_subscription.rs
AndreiEres: good question

AndreiEres: now we're checking it after the tool connects to the chain.

parachain-tracer/src/main.rs
```diff
@@ -87,6 +89,15 @@ pub(crate) struct ParachainTracerOptions {
 	/// Defines subscription mode
 	#[clap(flatten)]
 	collector_opts: CollectorOptions,
+	/// Turns on historical mode to trace parachains between specific blocks instead of following chain updates
```
sandreim, line 92:
```suggestion
	/// Run in historical mode to trace parachains between specific blocks instead of following live chain progress
```

PR #405, opened 2023-06-29
[WIP] Allow using env vars for rpc node url in tests

AndreiEres:
@alvicsam give it a try :-) 
We need to add RPC_NODE_URL to env vars


alvicsam:
I tried but it failed: the node needs several minutes to be ready even with `warp` sync.

Issue #381, opened 2023-06-05
Introspector fails after chain metadata update


AndreiEres:
As we use subxt, we are required to use a preloaded metadata. If the network's metadata changes, the introspector crashes. 

Currently, we check in the CI  for each PR if the metadata is up-to-date. This helps us identify the problem within a day or two, but we want a better solution.

The possible option is to create automated PRs to update the metadata. However, ideally, we would prefer to have no static metadata at all. Is this possible? 

Let's share and discuss ideas
@sandreim, @vstakhov, @lexnv, @prybalko

lexnv:
Updating the metadata with every runtime upgrade can be indeed a tedious task.

> If the network's metadata changes, the introspector crashes.

I wonder, does this happen with every runtime update?

I believe there are a few options to make this transition a bit less painful:
1. Instead of relying on downloaded subxt metadata, the subxt macro is able to fetch the metadata from a given URL

This still offers the same guarantees that are currently exposed by subxt, mainly compile-time safety to ensure that taking with the node is still possible. 
The change would require modifying all subxt macro invocations to point to the RPC endpoint, as opposed to loading the metadata from the disk.

https://github.com/paritytech/polkadot-introspector/blob/71193993c3d5b3e985f3380e8ee582bfcc8fc052/essentials/src/metadata.rs#L24-L33

For example, replacing the `runtime_metadata_path = "assets/rococo_metadata.scale"` with  `runtime_metadata_url = "wss://rpc.polkadot.io:443"`. 

When a runtime upgrade happens, the polkadot-interceptor binary would still need a binary update.

2. Using the dynamic API of subxt instead of relying on metadata entirely.

We expose the ability for users to create dynamic queries to the chain (see [dynamic storage query](https://github.com/paritytech/subxt/blob/1ad9a20ce2a156fae7c3b3d0fb4d408c5d07c374/subxt/examples/storage_fetch_dynamic.rs#L11-L13) for more details). 

This removes the need of having to manage the metadata as well.
A transition example would look similar to:

```rust
   // Static version:
   #[subxt::subxt(runtime_metadata_path = "../artifacts/polkadot_metadata_small.scale")]
   pub mod polkadot {}
   let storage_query = polkadot::storage().system().account(&account);

   // The dynamical version: 
   let storage_query =
        subxt::dynamic::storage("System", "Account", vec![Value::from_bytes(account)]);
``` 

[...]

PR #414, opened 2023-07-04
Use dynamic host configuration

AndreiEres:
Part of https://github.com/paritytech/polkadot-introspector/issues/381

We'd like to use dynamic API to not crash on every metadata update. That's a first try.


AndreiEres:
Right now we have incompatible kusama metadata but parachain-tracer works without crashing.

vstakhov (approved):
It looks very good and it solves the old problem of configuration mismatch!

Issue #178, opened 2022-12-13
Add telemetry backend support for dispute tracking


sandreim:
A nice enhancement to the dispute tracker is to be able to get additional information about dispute initiators, like hw information and OS. This would save a lot of time when trying to figure out who initiated disputes and why.

```
pub struct SysInfo {
	/// The exact CPU model.
	pub cpu: Option<String>,
	/// The total amount of memory, in bytes.
	pub memory: Option<u64>,
	/// The number of physical CPU cores.
	pub core_count: Option<u32>,
	/// The Linux kernel version.
	pub linux_kernel: Option<String>,
	/// The exact Linux distribution used.
	pub linux_distro: Option<String>,
	/// Whether the node's running under a virtual machine.
	pub is_virtual_machine: Option<bool>,
}
```

We just need to:
- add option to specify a telemetry server 
- when configured, connect and search for validators `network_id` which should by address we have from chain (IIRC when I looked at `sc_telemetry`). 
- fetch the additional information that is sent when the connection to telemetry backend is made, see below

```
pub struct ConnectionMessage {
	/// Node's name.
	pub name: String,
	/// Node's implementation.
	pub implementation: String,
	/// Node's version.
	pub version: String,
	/// Node's configuration.
	pub config: String,
	/// Node's chain.
	pub chain: String,
	/// Node's genesis hash.
	pub genesis_hash: String,
	/// Node is an authority.
	pub authority: bool,
	/// Node's startup time.
	pub startup_time: String,
	/// Node's network ID.
	pub network_id: String,

	/// Node's OS.
	pub target_os: String,

	/// Node's ISA.
	pub target_arch: String,

	/// Node's target platform ABI or libc.
	pub target_env: String,

	/// Node's software and hardware information.
	pub sysinfo: Option<SysInfo>,
}

PR #443, opened 2023-07-13
Add dispute initiators to dispute info

AndreiEres:
Part of https://github.com/paritytech/polkadot-introspector/issues/178

### DIFF
--- a/essentials/src/collector/mod.rs
+++ b/essentials/src/collector/mod.rs
@@ -780,7 +780,7 @@ impl Collector {
 		let now = get_unix_time_unwrap();
 		let para_id = candidate.parachain_id();
 		candidate.candidate_disputed = Some(CandidateDisputed { disputed: relay_block_number, concluded: None });
-		let initiator_indices = self.extract_dispute_initiators(dispute_event).await?;
+		let (initiator_indices, session_index) = self.extract_dispute_initiators(dispute_event).await?;
 		if let Some(to_websocket) = self.to_websocket.as_mut() {
 			to_websocket
 				.send(WebSocketUpdateEvent {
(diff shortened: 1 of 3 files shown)


essentials/src/collector/mod.rs
```diff
@@ -112,6 +113,7 @@ pub type CollectorStorageApi = ApiService<H256, CollectorPrefixType>;
 #[derive(Clone, Debug, Encode, Decode)]
 pub struct DisputeInfo {
 	pub initiated: <PolkadotConfig as subxt::Config>::Index,
+	pub initiators_idx: Vec<u32>,
```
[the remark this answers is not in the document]
vstakhov: We might also try to extract the validator address like we do for disputes tracking.

essentials/src/collector/mod.rs
```diff
@@ -781,7 +795,9 @@ impl Collector {
 		// Fill and write dispute info structure
 		let dispute_info = DisputeInfo {
 			dispute: dispute_event.clone(),
-			initiated: self.state.current_relay_chain_block_number,
+			initiated: relay_block_number,
+			initiator_indices,
+			session_index: self.state.current_session_index,
```
ordian, line 800: I would suggesting reading `session_index` from the `InherentData`, i.e. `extract_dispute_initiators` should return a tuple `(indices, session_index)`. Even if dispute is initiated in the current session, doesn't *necessarily* mean that validators indices are from the current session, they are from when the candidate was included, which may or may not be the current session.

parachain-tracer/src/progress.rs
```diff
@@ -245,6 +245,16 @@ impl Display for DisputesTracker {
 			},
 		}
 
+		if !self.initiators.is_empty() {
+			for (validator_idx, validator_address) in &self.initiators {
+				writeln!(
+					f,
+					"\t\t\t🚀 Validator initiated dispute: {}",
```
ordian, line 252: Not sure what's the best emoji for initiating a dispute, but I don't view it as a positive event :)
```suggestion
					"\t\t\t😠 Validator initiated dispute: {}",
```

ordian (approved):

vstakhov (approved):

PR #454, opened 2023-07-18
Use dynamic api for metadata-sensitive subxt calls

AndreiEres:
Related to https://github.com/paritytech/polkadot-introspector/issues/381

These three methods are metadata-sensitive. I made them use the dynamic API, adding some decoding rules.

### DIFF
--- a/essentials/src/api/dynamic.rs
+++ b/essentials/src/api/dynamic.rs
@@ -15,10 +15,10 @@ pub(crate) fn decode_dynamic_validator_groups(
 	raw_groups: &Value<u32>,
 ) -> Result<Vec<Vec<ValidatorIndex>>, SubxtWrapperError> {
 	let decoded_groups = decode_vector(raw_groups)?;
-	let mut groups = vec![];
+	let mut groups = Vec::with_capacity(decoded_groups.len());
 	for raw_group in decoded_groups.iter() {
 		let decoded_group = decode_vector(raw_group)?;
-		let mut group = vec![];
+		let mut group = Vec::with_capacity(decoded_group.len());
 		for raw_index in decoded_group.iter() {
 			group.push(ValidatorIndex(decode_u128_value(raw_index)? as u32));
 		}
@@ -32,7 +32,7 @@ pub(crate) fn decode_dynamic_availability_cores(
 	raw_cores: &Value<u32>,
 ) -> Result<Vec<Option<CoreOccupied>>, SubxtWrapperError> {
 	let decoded_cores = decode_vector(raw_cores)?;
-	let mut cores = vec![];
+	let mut cores = Vec::with_capacity(decoded_cores.len());
 	for raw_core in decoded_cores.iter() {
 		cores.push(decode_option(raw_core)?.map(|v| match decode_variant(v).unwrap().name.as_str() {
 			"Parachain" => CoreOccupied::Parachain,
@@ -45,7 +45,7 @@ pub(crate) fn decode_dynamic_availability_cores(
 
 pub(crate) fn decode_dynamic_scheduled_paras(raw_paras: &Value<u32>) -> Result<Vec<CoreAssignment>, SubxtWrapperError> {
 	let decoded_paras = decode_vector(raw_paras)?;
-	let mut paras = vec![];
+	let mut paras = Vec::with_capacity(decoded_paras.len());
 	for para in decoded_paras.iter() {
 		let core = CoreIndex(decode_u128_value(para.at("core").expect("Should be defined"))? as u32);
 		let para_id = Id(decode_u128_value(para.at("para_id").expect("Should be defined"))? as u32);
--- a/essentials/src/api/dynamic.rs
+++ b/essentials/src/api/dynamic.rs
@@ -1,0 +1,47 @@
+use super::subxt_wrapper::SubxtWrapperError::{self, DecodeDynamicError};
+use crate::metadata::polkadot_primitives;
+use subxt::{
+	dynamic::Value,
+	ext::scale_value::{Composite, Primitive, ValueDef},
+};
+
+pub(crate) fn decode_dynamic_validator_groups(
+	raw_groups: &Value<u32>,
+) -> Result<Vec<Vec<polkadot_primitives::ValidatorIndex>>, SubxtWrapperError> {
+	let decoded_groups = decode_validator_groups_value(raw_groups)?;
+	let mut groups = vec![];
+	for raw_group in decoded_groups.iter() {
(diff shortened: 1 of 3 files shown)


essentials/src/api/subxt_wrapper.rs
```diff
@@ -517,20 +519,26 @@ fn decode_extrinsic(data: &mut &[u8]) -> std::result::Result<SubxtCall, DecodeEx
 }
 
 async fn subxt_get_sheduled_paras(api: &OnlineClient<PolkadotConfig>, block_hash: H256) -> Result {
-	let addr = polkadot::storage().para_scheduler().scheduled();
-	let scheduled_paras = api.storage().at(block_hash).fetch(&addr).await?.unwrap_or_default();
-	Ok(Response::ScheduledParas(scheduled_paras))
+	let addr = subxt::dynamic::storage_root("ParaScheduler", "Scheduled");
+	let value = api.storage().at(block_hash).fetch(&addr).await?.unwrap().to_value()?;
```
lexnv, line 523: nit: `.await?.unwrap().to_value()?;` the `unwrap` here and similar below could fail if we stop providing the storage entry, would except make it more explicit?

essentials/src/api/dynamic.rs
```diff
@@ -0,0 +1,92 @@
...
+		}));
+	}
+
+	Ok(cores)
+}
+
+pub(crate) fn decode_dynamic_scheduled_paras(raw_paras: &Value<u32>) -> Result<Vec<CoreAssignment>, SubxtWrapperError> {
+	let decoded_paras = decode_vector(raw_paras)?;
+	let mut paras = vec![];
+	for para in decoded_paras.iter() {
```
vstakhov, line 49:
```suggestion
	for para in &decoded_paras {
```

essentials/src/api/subxt_wrapper.rs
vstakhov: If we return a Result, it might be even worth to convert that to a non-fatal error (especially for the mode with running on all heads)

essentials/src/api/subxt_wrapper.rs
```diff
@@ -517,20 +519,26 @@ fn decode_extrinsic(data: &mut &[u8]) -> std::result::Result<SubxtCall, DecodeEx
...
+	let cores = decode_dynamic_availability_cores(&value)?;
+
+	Ok(Response::OccupiedCores(cores))
 }
 
 async fn subxt_get_validator_groups(api: &OnlineClient<PolkadotConfig>, block_hash: H256) -> Result {
-	let addr = polkadot::storage().para_scheduler().validator_groups();
-	let groups = api.storage().at(block_hash).fetch(&addr).await?.unwrap_or_default();
+	let addr = subxt::dynamic::storage_root("ParaScheduler", "ValidatorGroups");
+	let value = api.storage().at(block_hash).fetch(&addr).await?.unwrap().to_value()?;
```
vstakhov, line 539: Same.

vstakhov (approved):
LGTM with a few nits.

essentials/src/api/subxt_wrapper.rs
AndreiEres: Done

essentials/src/api/dynamic.rs
```diff
@@ -0,0 +1,92 @@
...
+			_ => todo!("Add parathreads support"),
+		}));
+	}
+
+	Ok(cores)
+}
+
+pub(crate) fn decode_dynamic_scheduled_paras(raw_paras: &Value<u32>) -> Result<Vec<CoreAssignment>, SubxtWrapperError> {
+	let decoded_paras = decode_vector(raw_paras)?;
+	let mut paras = vec![];
```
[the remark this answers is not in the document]
AndreiEres: Done

essentials/src/api/dynamic.rs
AndreiEres: It doesn't work, because `decaded_paras` are `&Vec`, not `Vec`.

essentials/src/api/dynamic.rs
```diff
@@ -0,0 +1,92 @@
...
+	for para in decoded_paras.iter() {
+		let core = CoreIndex(decode_u128_value(para.at("core").expect("Should be defined"))? as u32);
+		let para_id = Id(decode_u128_value(para.at("para_id").expect("Should be defined"))? as u32);
+		let kind = match decode_variant(para.at("kind").expect("Should be defined"))?.name.as_str() {
```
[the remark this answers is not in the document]
AndreiEres: You're right, because it's dynamic, better to throw a non fatal error.

essentials/src/api/subxt_wrapper.rs
```diff
@@ -645,7 +662,11 @@ async fn subxt_get_hrmp_content(
 
 async fn subxt_get_host_configuration(api: &OnlineClient<PolkadotConfig>) -> Result {
 	let addr = subxt::dynamic::storage_root("Configuration", "ActiveConfig");
-	let value = api.storage().at_latest().await?.fetch(&addr).await?.unwrap().to_value()?;
+	let value = match api.storage().at_latest().await?.fetch(&addr).await? {
```
lexnv, line 665: nit: Do we want to fetch the `ActiveConfig` from the latest block, or from a dedicated block hash to keep in line with the other functions?

essentials/src/api/dynamic.rs
```diff
@@ -0,0 +1,92 @@
...
+	raw_cores: &Value<u32>,
+) -> Result<Vec<Option<CoreOccupied>>, SubxtWrapperError> {
+	let decoded_cores = decode_vector(raw_cores)?;
+	let mut cores = vec![];
+	for raw_core in decoded_cores.iter() {
+		cores.push(decode_option(raw_core)?.map(|v| match decode_variant(v).unwrap().name.as_str() {
+			"Parachain" => CoreOccupied::Parachain,
+			_ => todo!("Add parathreads support"),
+		}));
+	}
```
lexnv, line 41: nit:
```suggestion
	let cores = decoded_cores.iter().map(|raw_core| {
	       Ok(decode_option(raw_core)?.map(|v| match decode_variant(v).unwrap().name.as_str() {
			"Parachain" => CoreOccupied::Parachain,
			_ => todo!("Add parathreads support"),
		})))
	}).collect::<Result<_,_>()?;
```

essentials/src/api/subxt_wrapper.rs
AndreiEres: In that particular case we're good with the latest.

PR #462, opened 2023-07-21
Move out responsibility for the launching from EventStream

AndreiEres:
This way we can run more than one subscription, like to add telemetry feed to parachain-tracer

### DIFF
--- a/block-time/src/main.rs
+++ b/block-time/src/main.rs
@@ -21,7 +21,7 @@ use crossterm::{
 	terminal::{Clear, ClearType},
 	QueueableCommand,
 };
-use log::{debug, error, info, warn};
+use log::{debug, info, warn};
 use polkadot_introspector_essentials::{
 	api::ApiService,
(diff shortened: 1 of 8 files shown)


block-time/src/main.rs
```diff
@@ -400,16 +386,18 @@ fn register_metric(registry: &Registry) -> HistogramVec {
...
-
-	match BlockTimeMonitor::new(opts, block_time_consumer_init)?.run().await {
-		Ok(futures) =>
-			core.run(futures, shutdown_tx.clone(), tokio::spawn(init::on_shutdown(shutdown_tx.clone())))
-				.await?,
-		Err(err) => error!("FATAL: cannot start block time monitor: {}", err),
-	}
+
+	let monitor = BlockTimeMonitor::new(opts.clone())?;
+	let shutdown_tx = init::init_shutdown();
```
vstakhov, line 391: 👍

vstakhov (approved):
Nice work!

PR #467, opened 2023-07-26
Add relay skipped slots metric

AndreiEres:
When a candidate misses its slot due to a delayed relay chain block, it’s not the candidate's fault. To ensure accurate measurements we propose to add a new metric to track "relay skipped slots". Technically, the relay chain doesn't have "skipped slots" so the more appropriate name could be `relay_relative_block_time`.

New metric represents an actual block time divided by a standard block time
6000ms -> 6000/6000 = 1
12000ms -> 12000/6000 = 2

To take into account a max drift of 3000ms we round results
8999ms -> 8999/6000 = 1
9000ms -> 9000/6000 = 2

### DIFF
--- a/parachain-tracer/src/prometheus.rs
+++ b/parachain-tracer/src/prometheus.rs
@@ -58,6 +58,8 @@ struct MetricsInner {
 	disputes_stats: DisputesMetrics,
 	/// Block time measurements for relay parent blocks
 	relay_block_times: HistogramVec,
+	/// Relative time measurements (in standard blocks) for relay parent blocks
+	relay_relative_block_times: HistogramVec,
 	/// Number of slow availability events.
 	slow_avail_count: IntCounterVec,
 	/// Number of low bitfield propagation events.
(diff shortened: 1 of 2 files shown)


vstakhov:
Hum, why do we need another metric if it can be derived from an existing metric by just dividing by `6.0` which could be done in `promql` itself when making a query?

parachain-tracer/src/prometheus.rs
```diff
@@ -88,11 +90,19 @@ impl Metrics {
 	}
 
 	pub(crate) fn on_block(&self, time: f64, para_id: u32) {
+		const STANDARD_BLOCK_TIME: f64 = 6.0;
```
sandreim, line 93: This should be moved into some constants file along with others.

parachain-tracer/src/prometheus.rs
```diff
@@ -88,11 +90,19 @@ impl Metrics {
...
 		if let Some(metrics) = &self.0 {
 			metrics
 				.relay_block_times
 				.with_label_values(&[&para_id.to_string()[..]])
```
sandreim, line 97: Not really related to this PR, but we should only compute relay chain block time only once. Currently we do this for every para.

sandreim (approved):

PR #476, opened 2023-08-01
Add support of the Westend chain

vstakhov:
In fact, we should only mention these 4 networks as officially supported. Versi is the internal network and it can have whatever runtime we want for our internal testing.


vstakhov:
We might probably get rid of features soon, so this PR is likely not needed.

Issue #387, opened 2023-06-09
Simplify or remove a hack with the ParaInherent extrinsic decoding


AndreiEres:
Since https://github.com/paritytech/subxt/pull/929 we don't need it

AndreiEres:
I'm not sure that we can fix it right now, because we use different data types. I'd rather close it for now.

PR #477, opened 2023-08-01
Decode parainherent via subxt

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/387

Changes also should make data decoding more independent of metadata changes.

### DIFF
--- a/essentials/src/api/subxt_wrapper.rs
+++ b/essentials/src/api/subxt_wrapper.rs
@@ -19,10 +19,9 @@ use super::dynamic::decode_dynamic_validator_groups;
 use crate::{
 	api::dynamic::{decode_dynamic_availability_cores, decode_dynamic_scheduled_paras},
 	metadata::{polkadot, polkadot_primitives},
-	types::{AccountId32, BlockNumber, SessionKeys, SubxtCall, Timestamp, H256},
+	types::{AccountId32, BlockNumber, SessionKeys, Timestamp, H256},
 	utils::{Retry, RetryOptions},
 };
-use codec::Decode;
 use log::error;
 use std::{
 	collections::{hash_map::HashMap, BTreeMap},
@@ -52,7 +51,7 @@ pub enum RequestType {
 	/// Get block events.
 	GetEvents(<PolkadotConfig as subxt::Config>::Hash),
 	/// Extract the `ParaInherentData` from a given block.
-	ExtractParaInherent(subxt::rpc::types::ChainBlock<PolkadotConfig>),
+	ExtractParaInherent(subxt::blocks::Block<PolkadotConfig, OnlineClient<PolkadotConfig>>),
 	/// Get the availability core scheduling information at a given block.
 	GetScheduledParas(<PolkadotConfig as subxt::Config>::Hash),
 	/// Get occupied core information at a given block.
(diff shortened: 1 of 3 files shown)


AndreiEres:
@lexnv Do you have any idea how to do it better?

essentials/src/api/subxt_wrapper.rs
```diff
@@ -473,7 +471,11 @@ async fn subxt_get_block_ts(api: &OnlineClient<PolkadotConfig>, hash: H256) -> R
 }
 
 async fn subxt_get_block(api: &OnlineClient<PolkadotConfig>, maybe_hash: Option<H256>) -> Result {
-	Ok(Response::MaybeBlock(api.rpc().block(maybe_hash).await?.map(|response| response.block)))
+	let block = match maybe_hash {
+		Some(hash) => api.blocks().at(hash).await?,
+		None => api.blocks().at_latest().await?,
+	};
```
lexnv, line 477: nit:

```suggestion
	let block = api.blocks().at_or_latest(maybe_hash).await?;
```

essentials/src/api/subxt_wrapper.rs
```diff
@@ -690,16 +671,28 @@ async fn subxt_unpin_chain_head(api: &OnlineClient<PolkadotConfig>, sub_id: &str
...
+	let ex = body
+		.extrinsics()
+		.iter()
+		.take(2)
+		.last()
+		.expect("`ParaInherent` data is always at index #1")
+		.expect("`ParaInherent` data must exist");
+
+	let data = match ex
+		.as_root_extrinsic::<polkadot::Call>()
```
lexnv, line 687: nit: This works as intended, we also have the ability to decode directly to a pallet type

```
let data =  match ex.as_extrinsic<polkadot::para_inherent::calls::types::Enter>() 
```

We have a testing example [here](https://github.com/paritytech/subxt/blob/fd853b9b72b5335948d0a54c33cf3a05062caee4/testing/integration-tests/src/full_client/blocks/mod.rs#L170-L173), but again totally up to you since it works

essentials/src/api/subxt_wrapper.rs
```diff
@@ -497,27 +499,6 @@ pub enum DecodeExtrinsicError {
 	CodecError(codec::Error),
 }
 
-/// Decode a Polkadot emitted extrinsic from provided bytes.
-fn decode_extrinsic(data: &mut &[u8]) -> std::result::Result<SubxtCall, DecodeExtrinsicError> {
```
lexnv, line 501: Happy to see this removed! Nice one!

lexnv (approved):
Solid job here! Thanks for relying on subxt to decode extrinsics instead of doing it manually, it simplifies the code quite a bit!

essentials/src/api/subxt_wrapper.rs
AndreiEres: I'd love to use `at_or_latest` but it is a private method :-)

lexnv: Ahh I see, we removed in favor of `at` and `at_latest`

vstakhov (approved):
I'm glad if it allows us to get rid of features eventually.

PR #486, opened 2023-08-08
Show backing context size in the tracker

vstakhov:
This should make AB more visible and address #479.


parachain-tracer/src/tracker.rs
```diff
@@ -717,10 +717,29 @@ impl SubxtTracker {
 						self.current_candidate.max_av_bits,
 					));
 				}
-				self.stats.on_included(relay_block_number, self.last_included_block);
+
+				// Extract stored candidate from the collector if any
+				let stored_candidate = self
+					.api
+					.storage()
+					.storage_read_prefixed(CollectorPrefixType::Candidate(self.para_id), candidate_hash)
```
AndreiEres, line 725: fmt?

AndreiEres (approved):

PR #487, opened 2023-08-09
Prevent breaking changes after on-demand parachains released

AndreiEres:
On-demand Parachains shows up with some type changes. In that PR I added some new types, which could be not exact as the original ones, but they're enough to make tracer work. We need to replace them with original ones as they're released in v5.

### DIFF
--- a/essentials/src/api/dynamic.rs
+++ b/essentials/src/api/dynamic.rs
@@ -5,9 +5,9 @@ use crate::{
 			polkadot_parachain::primitives::Id,
 			polkadot_runtime_parachains::scheduler::{AssignmentKind, CoreAssignment},
 		},
-		polkadot_primitives::{CoreIndex, CoreOccupied, GroupIndex, ValidatorIndex},
+		polkadot_primitives::{CoreIndex, GroupIndex, ValidatorIndex},
 	},
-	types::{Assignment, BlockNumber, ClaimQueue, ParasEntry},
+	types::{Assignment, BlockNumber, ClaimQueue, CoreOccupied, ParasEntry},
 };
 use log::error;
 use std::collections::{BTreeMap, VecDeque};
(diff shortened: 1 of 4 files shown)


essentials/src/api/dynamic.rs
```diff
@@ -71,6 +75,39 @@ pub(crate) fn decode_dynamic_scheduled_paras(raw_paras: &Value<u32>) -> Result<V
 	Ok(paras)
 }
 
+pub(crate) fn decode_dynamic_claim_queue(raw: &Value<u32>) -> Result<ClaimQueue, SubxtWrapperError> {
+	let decoded_btree_map = decode_unnamed_composite(raw)?;
+	let decoded_values = decode_unnamed_composite(decoded_btree_map.first().expect("Must contain values"))?;
+	let mut claim_queue: ClaimQueue = BTreeMap::new();
+	for value in decoded_values {
+		let decoded_tuple = decode_unnamed_composite(value)?;
+		let raw_core = decoded_tuple.first().expect("Must contain core");
```
sandreim, line 84: we should just return error and not panic when decoding values.

AndreiEres: That's a good remark. Thank you.

vstakhov (approved):

essentials/src/api/dynamic.rs
```diff
@@ -71,6 +82,47 @@ pub(crate) fn decode_dynamic_scheduled_paras(raw_paras: &Value<u32>) -> Result<V
 	Ok(paras)
 }
 
+pub(crate) fn decode_dynamic_claim_queue(raw: &Value<u32>) -> Result<ClaimQueue, SubxtWrapperError> {
```
sandreim, line 85: TBH, we could just drop the `dynamic` in the fn names.

essentials/src/types.rs
```diff
@@ -31,3 +32,35 @@ pub type AccountId32 = utils::AccountId32;
...
+	pub para_id: u32,
+}
+
+// TODO: Take it from runtime types v5
+/// Temporary abstraction to cover core state until v5 types are released
+#[derive(Debug)]
+pub enum CoreOccupied {
+	/// The core is not occupied.
+	Free,
+	/// A paras.
```
sandreim, line 64: The naming is a bit confusing, Is it occupied by a parachain or many ?

parachain-tracer/src/tracker.rs
```diff
@@ -521,19 +524,59 @@ impl SubxtTracker {
...
 	}
 
-	fn update_core_assignment(
+	async fn get_core_assignments_via_scheduled_paras(
 		&mut self,
-		core_assignments: Vec<polkadot::runtime_types::polkadot_runtime_parachains::scheduler::CoreAssignment>,
-	) {
-		if let Some(index) = core_assignments
+		block_hash: H256,
+	) -> color_eyre::Result<Option<HashMap<u32, Vec<u32>>>> {
```
sandreim, line 530: I suggest we get rid of the Option. Just match errors at upper level where we can decide  to try to fetch `ParaScheduler,ClaimQueue` instead

sandreim (approved):

essentials/src/types.rs
AndreiEres: It's an original name from Polkadot (on-demand pallet branch)

AndreiEres: It's an original name from Polkadot (on-demand pallet branch)

Issue #494, opened 2023-08-15
Support async backing changes


sandreim:
There are a few things changed in the async backing branch https://github.com/paritytech/polkadot/tree/rh-async-backing-feature that is going to merge soon.

While trying tracer on Versi I hit this error:
```
Host configuration for wss://versi-rpc-node-0.parity-versi.parity.io:443:
        👀 Max validators: 0 / 5 per core
        👍 Needed approvals: 30
        🥔 No show slots: 2
        ⏳ Delay tranches: 89
Parachain Tracer will trace parachain(s) 501 on wss://versi-rpc-node-0.parity-versi.parity.io:443
-----------------------------------------------------------------------
[2023-08-15T15:45:09.328137Z ERROR polkadot_parachain_tracer] error occurred when processing block 0xc8c6…a2c6: 
       0: decode dynamic value error: expected `.group_idx`, got { core: (0), para_id: (501), kind: Parachain () }
    
    Location:
       /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/convert/mod.rs:727
```

PR #509, opened 2023-08-22
Support new types after the async backing merge

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/494

### DIFF
--- a/essentials/src/api/dynamic.rs
+++ b/essentials/src/api/dynamic.rs
@@ -2,12 +2,11 @@ use super::subxt_wrapper::SubxtWrapperError::{self, DecodeDynamicError};
 use crate::{
 	metadata::{
 		polkadot::runtime_types::{
-			polkadot_parachain::primitives::Id,
-			polkadot_runtime_parachains::scheduler::{AssignmentKind, CoreAssignment},
+			polkadot_parachain::primitives::Id, polkadot_runtime_parachains::scheduler::AssignmentKind,
 		},
-		polkadot_primitives::{CoreIndex, GroupIndex, ValidatorIndex},
+		polkadot_primitives::{CoreIndex, ValidatorIndex},
 	},
-	types::{Assignment, BlockNumber, ClaimQueue, CoreOccupied, ParasEntry},
+	types::{Assignment, BlockNumber, ClaimQueue, CoreAssignment, CoreOccupied, ParasEntry},
 };
 use log::error;
 use std::collections::{BTreeMap, VecDeque};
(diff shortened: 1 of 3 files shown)


essentials/src/api/dynamic.rs
```diff
@@ -113,10 +111,10 @@ fn decode_paras_entry_option(raw: &Value<u32>) -> Result<Option<ParasEntry>, Sub
 fn decode_paras_entry(raw: &Value<u32>) -> Result<ParasEntry, SubxtWrapperError> {
 	let para_id = decode_composite_u128_value(value_at("para_id", value_at("assignment", raw)?)?)? as u32;
 	let assignment = Assignment { para_id };
-	let retries = decode_u128_value(value_at("retries", raw)?)? as u32;
+	let availability_timeouts = decode_u128_value(value_at("availability_timeouts", raw)?)? as u32;
```
vstakhov, line 114: I think we need to check both variants and decode what we can here.

essentials/src/types.rs
```diff
@@ -35,14 +41,25 @@ pub type SubxtCall = runtime::RuntimeCall;
...
 // TODO: Take it from runtime types v5
 /// Polkadot v5 ParasEntry type
 #[derive(Debug)]
 pub struct ParasEntry {
 	/// The `Assignment`
 	pub assignment: Assignment,
-	/// Number of times this has been retried.
-	pub retries: u32,
+	/// The number of times the entry has timed out in availability.
+	pub availability_timeouts: u32,
```
vstakhov, line 62: The meaning of these fields seems to be quite similar, so we should probably use it for the both elements returned from the runtime.

essentials/src/types.rs
```diff
@@ -35,14 +41,25 @@ pub type SubxtCall = runtime::RuntimeCall;
 
 pub type ClaimQueue = BTreeMap<u32, VecDeque<Option<ParasEntry>>>;
 
+// TODO: Take it from runtime types v5
+/// How a free core is scheduled to be assigned.
+pub struct CoreAssignment {
```
antonva, line 46: Just a heads up, on master, this type has changed a bit. See here: https://github.com/paritytech/polkadot/blob/759fe213acd6071b4a2bfebcef08be8a5f2ecc15/runtime/parachains/src/scheduler/common.rs#L79

essentials/src/types.rs
antonva: This change was made to give the field a more descriptive name due to ambiguity.

essentials/src/api/dynamic.rs
antonva: There hasn't been a runtime released that uses `ParasEntry` yet. I think it would be safe to assume that the `retries` field will never be seen.

antonva (commented):
Looks correct to me.

essentials/src/types.rs
AndreiEres: Thank you. Here it's ok, we need it to support "old", not updated to v5 blockchains.

vstakhov (approved):

Issue #483, opened 2023-08-04
Introspector closed channels but didn't reload after network error


AndreiEres:
```
[2023-08-02T12:01:55.513964Z ERROR polkadot_introspector_essentials::collector] collector service could not process event: subxt error: Rpc error: RPC error: RPC call failed: ErrorObject { code: ServerError(-32000), message: "Client error: UnknownBlock: State already discarded for 0x2810d818a55c8ab39e6a4ae7b064350d28e03b629b02cbb5d7b2e25093b058a5", data: None }
[2023-08-02T12:01:55.515133Z INFO  polkadot_parachain_tracer] Received termination event, 38 trackers will be terminated, 38 futures are pending
[2023-08-02T12:01:55.515252Z INFO  polkadot_parachain_tracer] Input channel has been closed
```

PR #511, opened 2023-08-23
Take down the tracer on the collector termination

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/483

We should take down the entire tracer on the collector termination to allow k8s to restart the pod.

### DIFF
--- a/essentials/src/collector/mod.rs
+++ b/essentials/src/collector/mod.rs
@@ -161,8 +161,17 @@ pub enum CollectorUpdateEvent {
 	NewHead(NewHeadEvent),
 	/// Occurs on a new session
 	NewSession(u32),
-	/// Occurs when collector is disconnected and is about to terminate, contains error code
-	Termination(i32),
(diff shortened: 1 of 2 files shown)


essentials/src/collector/mod.rs
```diff
@@ -161,8 +161,8 @@ pub enum CollectorUpdateEvent {
 	NewHead(NewHeadEvent),
 	/// Occurs on a new session
 	NewSession(u32),
-	/// Occurs when collector is disconnected and is about to terminate
-	Termination,
+	/// Occurs when collector is disconnected and is about to terminate, contains error code
+	Termination(i32),
```
vstakhov, line 165: I think it should be something like 
```rust
enum CollectorTermination {
  Normal,
  Abnormal(i32, String)
}
```

This will distinguish normal from abnormal explicitly and let collector pass extra information about termination reason.

vstakhov (approved):
Nice!

Issue #375, opened 2023-05-26
Support for on-demand parachains


sandreim:
The on-demand parachain feature is currently under development. As soon as we have it merged in the Rococo runtime, we should look into implementing:
- claim queue monitoring - https://github.com/paritytech/polkadot/blob/alex/parathreads_review/runtime/parachains/src/scheduler.rs#L122
- bidding extrinsics
- latency between bid, claim and actual block production

vstakhov:
The tracker code should be able to support basic on-demand parachains tracking (with inactivity timeout). The rest is indeed would be useful to add.

AndreiEres:
With @antonva we decided that it would be useful to track `OnDemandOrderPlaced`

AndreiEres:
- [x] **claim queue monitoring**
Monitoring the claim queue instead of deprecated `scheduled` state call to check cores occupancy.

- [ ] **bidding extrinsics**
Tracking of event `OnDemandOrderPlaced { para_id: ParaId, spot_price: BalanceOf<T> }`.

- [ ] **latency between bid, claim and actual block production**
Tracking of a length in blocks between relay block with event `OnDemandOrderPlaced` and relay block where parablock from the corresponding parachain was included.

PR #518, opened 2023-08-24
Track on-demand latency

AndreiEres:
Closes https://github.com/paritytech/polkadot-introspector/issues/375
Added tracking of a length in blocks and seconds between a relay block with event `OnDemandOrderPlaced` and a relay block where candidate from the corresponding parachain was included.

### DIFF
--- a/parachain-tracer/src/prometheus.rs
+++ b/parachain-tracer/src/prometheus.rs
@@ -75,6 +75,10 @@ struct MetricsInner {
 	para_block_times_sec: HistogramVec,
 	/// Parachain's on-demand orders
 	para_on_demand_orders: GaugeVec,
+	/// Latency between ordering a slot by a parachain and its last included candidate in relay blocks
+	para_on_demand_delay: GaugeVec,
+	/// Latency between ordering a slot by a parachain and its last included candidate in seconds
+	para_on_demand_delay_sec: GaugeVec,
 	/// Finality lag
 	finality_lag: Gauge,
 }
@@ -197,6 +201,7 @@ impl Metrics {
 		}
 	}
 
+	/// Update on-demand orders
 	pub(crate) fn on_on_demand_order(&self, order: &OnDemandOrder) {
 		if let Some(metrics) = &self.0 {
 			let para_str: String = order.para_id.to_string();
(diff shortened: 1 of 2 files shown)


parachain-tracer/src/prometheus.rs
```diff
@@ -75,6 +75,10 @@ struct MetricsInner {
 	para_block_times_sec: HistogramVec,
 	/// Parachain's on-demand orders
 	para_on_demand_orders: GaugeVec,
+	/// Latency between ordering a slot by a parachain and its last included candidate in relay blocks
+	para_on_demand_delay: GaugeVec,
+	/// Latency between ordering a slot by a parachain and its last included candidate in seconds
```
sandreim, line 80: The description is not accurate, the candidate is considered included when it becomes availability, in our case the latency should measure the time to being backed.

parachain-tracer/src/tracker.rs
```diff
@@ -408,6 +420,17 @@ impl SubxtTracker {
 			.storage_read_prefixed(CollectorPrefixType::OnDemandOrder(self.para_id), block_hash)
 			.await
 			.map(|v| v.into_inner::<OnDemandOrder>().unwrap());
+
+		if self.on_demand_order.is_none() {
```
vstakhov, line 424: I'd prefer more canonical `self.on_demand_order.map`.

vstakhov (approved):

parachain-tracer/src/tracker.rs
AndreiEres: What do you mean in that case?

parachain-tracer/src/prometheus.rs
```diff
@@ -207,6 +212,28 @@ impl Metrics {
 		}
 	}
 
+	/// Update on-demand latency in blocks
+	pub(crate) fn on_on_demand_delay(&self, delay_blocks: u32, para_id: u32) {
```
sandreim, line 216: These functions could use better naming:

```suggestion
	pub(crate) fn handle_on_demand_delay(&self, delay_blocks: u32, para_id: u32) {
```

parachain-tracer/src/prometheus.rs
```diff
@@ -339,6 +366,20 @@ fn register_metrics(registry: &Registry) -> Result<Metrics> {
 			)?,
 			registry,
 		)?,
+		para_on_demand_delay: prometheus_endpoint::register(
+			GaugeVec::new(
+				Opts::new("pc_para_on_demand_delay", "Latency between ordering a slot by a parachain and its last backed candidate in relay blocks"),
```
sandreim, line 371:
```suggestion
				Opts::new("pc_para_on_demand_delay", "Latency (in relay chain blocks) between when the parachain orders a core and when first candidate is backed on that core."),
```

parachain-tracer/src/tracker.rs
```diff
@@ -304,6 +308,15 @@ impl ParachainBlockTracker for SubxtTracker {
 					}
 					self.stats.on_backed();
 					metrics.on_backed(self.para_id);
+					if let Some(last_on_demand_order_block) = self.last_on_demand_order_block {
+						metrics.on_on_demand_delay(
+							relay_block_number.saturating_sub(last_on_demand_order_block),
+							self.para_id,
+						)
+					}
+					if let Some(diff) = self.get_time_diff(self.current_relay_block_ts, self.last_on_demand_order_ts) {
```
sandreim, line 317: This seems to not count the latency just for the first block. It is possible that the core might be assigned to same parachain for multiple blocks so there will be an increase in latency as blocks are built.

parachain-tracer/src/tracker.rs
```diff
@@ -304,6 +310,13 @@ impl ParachainBlockTracker for SubxtTracker {
 					}
 					self.stats.on_backed();
 					metrics.on_backed(self.para_id);
+
+					if self.handle_on_demand_delay("backed", metrics) {
+						self.on_demand_order_block = None;
```
sandreim, line 315: this could be moved in `handle_on_demand_delay`. Ideally there is only one place where you mutate these `on_demand_*` variables, otherwise things get really messy. 

Overall readability is not that great as we add more functionality.

I'd love to see a followup refactoring PR.

sandreim (approved):

Issue #545, opened 2023-09-14
Consider empty `ClaimQueue` on session boundaries


AndreiEres:
Currently, we assume empty responses from dynamic API as an error, which is not true with `ClaimQueue`. It's empty after a session change, so the introspector tries to fetch a new queue in a loop and closes itself after a particular number of attempts. We should consider the fact that the queue might be `None` on session boundaries.

PR #548, opened 2023-09-21
Accept None from dynamic storage as a valid response

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/545

Responses from dynamic storage contain optional value. `None` means that there is no value in the storage. It's not an error. For example, `ClaimQueue` may be `None` after a session change.

### DIFF
--- a/essentials/src/api/subxt_wrapper.rs
+++ b/essentials/src/api/subxt_wrapper.rs
@@ -528,7 +528,7 @@ async fn fetch_dynamic_storage(
 		.fetch(&subxt::dynamic::storage_root(pallet_name, entry_name))
 		.await?
 	{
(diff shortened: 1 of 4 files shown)


vstakhov:
I might be wrong here, but cannot we use the already existing `Result` for that purposes? We can split error type into 2 categories - critical errors and `FYI` errors.

AndreiEres:
> I might be wrong here, but cannot we use the already existing `Result` for that purposes? We can split error type into 2 categories - critical errors and `FYI` errors.

It is our current approach, which I think is wrong. I don't think that `None` should be considered as error. It's just a response without value because we have nothing to read from the storage.

PR #567, opened 2023-10-10
Update subxt

AndreiEres:
Updates `subxt` to the last version

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,52 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "aead"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
+name = "aes"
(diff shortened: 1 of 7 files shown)


vstakhov (approved):

lexnv (approved):
LGMT!

Your approach to constructing `LegacyRPCMethods` and `UnstableRPCMethods` should work fine. The advantage of exposing these structures is that you can be explicit in calling certain RPC methods.

Leaving some thoughts about the other possible approaches for solving this. However, that is entirely up to you, since I'm not 100% familiar with the code / future use-cases.
 
The downside of exposing the RPC layer directly is that you'll still need to manage the low-level `chainHead` subscription, unpinning blocks and re-establishing connectivity when receiving the `chainHead::stop::Event`.
Subxt offers support for those operations for free by constructing the client like so:

```rust
let (backend, mut driver) = unstable::UnstableBackend::builder().build(rpc_client);

    // The unstable backend needs driving:
    tokio::spawn(async move {
        use futures::StreamExt;
        while let Some(val) = driver.next().await {
            if let Err(e) = val {
                eprintln!("Error driving unstable backend: {e}");
                break;
            }
        }
    });

    let client = OnlineClient::from_backend(Arc::new(backend))
        .await
        .map_err(|e| format!("Cannot construct OnlineClient from backend: {e}"))?;
```

Then, you could use the `subxt::client` API as normally, and depending on the backend it will make calls to:
- RPC V2 `chainHead` when the unstable backend is constructed
- RPC V1 legacy when the legacy  backend is constructed

I think that will simplify the code a bit, but you'll lose the granularity of being able to call RPC methods explicitly. Again the code looks good, totally up to you! 👍

PR #575, opened 2023-10-17
Cache repeating rpc queries in the collector

AndreiEres:
### Done
1. Removed unused RPC queries, see `essentials/src/api/subxt_wrapper.rs`.
1. Common for all parachains queries cached in the collector (ts, occupied_cores, backing_groups, core_assignments).

### Results

Tested on Polkadot, in historical mode, between blocks 17762300-17762310.

#### Number of RPC requests
|  | before | after |
|--------|--------|--------|
| For 2 paras | 192 | 164 |
| For all (38) paras | 1571 | 722 |

With a large number of parachains on a given interval, the number of queries decreased by about half.

### DIFF
--- a/parachain-tracer/src/tracker_storage.rs
+++ b/parachain-tracer/src/tracker_storage.rs
@@ -17,7 +17,7 @@
 use polkadot_introspector_essentials::{
 	api::{storage::RequestExecutor, subxt_wrapper::InherentData},
 	collector::{candidate_record::CandidateRecord, CollectorPrefixType, DisputeInfo},
-	types::{AccountId32, OnDemandOrder, H256},
+	types::{AccountId32, OnDemandOrder, Timestamp, H256},
 };
 use subxt::config::{substrate::BlakeTwo256, Hasher};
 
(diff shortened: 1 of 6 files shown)


parachain-tracer/src/tracker_storage.rs
```diff
@@ -71,9 +73,37 @@ impl TrackerStorage {
...
-	pub async fn candidate(&self, candidate_hash: H256) -> Option<CandidateRecord> {
+	pub async fn candidate(&self, block_hash: H256) -> Option<CandidateRecord> {
 		self.storage
-			.storage_read_prefixed(CollectorPrefixType::Candidate(self.para_id), candidate_hash)
+			.storage_read_prefixed(CollectorPrefixType::Candidate(self.para_id), block_hash)
+			.await
+			.map(|v| v.into_inner().unwrap())
+	}
+
+	pub async fn block_timestamp(&self, block_hash: H256) -> Option<Timestamp> {
```
vstakhov, line 83: Please also add some documentation if possible. E.g. I would really like to see what is the meaning of `block_hash` here.

vstakhov (approved):
LGTM 👍

PR #580, opened 2023-10-20
Add light client support

AndreiEres:
Last version of subxt supports light client via smoldot. It can help us to reduce costs on running additional rpc nodes.

Optional light client added to parachain tracer. Other tools use only online (old) client. 

Currently, light client seems not stable to work with introspector. But it can be solved in subxt and smoldot crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ parity-scale-codec = { version = "3.6.5", default-features = false, features = [
 color-eyre = "0.6.2"
 colored = "2.0.4"
 crossterm = "0.26.1"
+dyn-clone = "1.0.14"
 env_logger = "0.10.1"
 erased-serde = "0.3.31"
 futures = "0.3.29"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ serde_json = "1.0.108"
 serde_urlencoded = "0.7.1"
 snap = "1.1.0"
 strum = { version = "0.25.0", features = ["derive"] }
-subxt = { default-features = false, features = ["jsonrpsee", "native"], version = "0.32.1" }
+subxt = { default-features = false, features = ["jsonrpsee", "native", "unstable-light-client"], version = "0.32.1" }
 base64 = "0.21.5"
 thiserror = "1.0.50"
 time = { version = "0.3.30", features = ["formatting"] }
--- a/essentials/src/api/api_client.rs
+++ b/essentials/src/api/api_client.rs
@@ -15,47 +15,55 @@
 // along with polkadot-introspector.  If not, see <http://www.gnu.org/licenses/>.
 //
 
-use crate::types::{BlockNumber, Header, H256};
+use crate::{
+	metadata::polkadot,
+	types::{AccountId32, BlockNumber, Header, InherentData, SessionKeys, SubxtHrmpChannel, Timestamp, H256},
+};
 use dyn_clone::DynClone;
+use std::collections::BTreeMap;
 use subxt::{
 	backend::{
 		legacy::{rpc_methods::NumberOrHex, LegacyRpcMethods},
 		rpc::RpcClient,
 		StreamOf,
 	},
-	blocks::{BlockRef, BlocksClient},
+	blocks::{Block, BlockRef, BlocksClient},
 	client::{LightClient, OnlineClientT},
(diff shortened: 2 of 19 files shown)


essentials/src/api/subxt_wrapper.rs
```diff
@@ -421,56 +417,66 @@ impl RequestExecutor {
 	}
 }
 
+/// How to subscribe to subxt blocks
+#[derive(strum::Display, Debug, Clone, Copy, ValueEnum, Default)]
+pub enum ApiClientMode {
+	#[default]
+	Online,
```
sandreim, line 424:
```suggestion
	RPC,
```

essentials/src/api/api_client.rs
```diff
@@ -15,66 +15,249 @@
...
+
+	Ok(ApiClient { client, legacy_rpc_methods })
+}
+
+pub async fn build_light_client(url: &str) -> Result<ApiClient<LightClient<PolkadotConfig>>, String> {
+	let rpc_client = RpcClient::from_url(url)
+		.await
+		.map_err(|e| format!("Cannot construct RPC client: {e}"))?;
+	let client = LightClient::builder()
+		.build_from_url(url)
```
sandreim, line 257: We still need an RPC endpoint here, why not use `build` from chainspec ?

AndreiEres: We can, but why it's better? We need only one call to download the chainspec. It's easy to connect to other networks. Otherwise, we need to have prepared a chainspecs for every particular network.

sandreim: The point was to not need an RPC node at all. Once the light client works fully, we should only use RPC to get a chainspec

AndreiEres: Unfortunately, chainspec changes over the time: fields `babeFinalizedBlockWeight`. `finalizedBlockHeader`, `grandpaAuthoritySet`. So it's better to fetch the fresh one when we connect.

sandreim: I see,  let’s download it for now

essentials/src/api/api_client.rs
```diff
@@ -15,66 +15,249 @@
...
+
+	async fn extract_parainherent(&self, maybe_hash: Option<H256>) -> Result<InherentData, subxt::Error> {
+		let block = self.block_at(maybe_hash).await?;
+		let ex = block
+			.extrinsics()
+			.await?
+			.iter()
+			.take(2)
+			.last()
+			.expect("`ParaInherent` data is always at index #1")
```
sandreim, line 213: I think we should not panic and return the error.

Cargo.toml
```diff
@@ -48,7 +49,7 @@ serde_json = "1.0.107"
 serde_urlencoded = "0.7.1"
 snap = "1.1.0"
 strum = { version = "0.25.0", features = ["derive"] }
-subxt = { default-features = false, features = ["jsonrpsee", "native"], version = "0.32.1" }
+subxt = { default-features = false, features = ["jsonrpsee", "native", "unstable-light-client"], version = "0.32.1" }
```
sandreim, line 52: How "stable" is the light client support currently ?

essentials/src/api/api_client.rs
```diff
@@ -15,66 +15,249 @@
...
+	dynamic::Value,
+	events::{Events, EventsClient},
 	storage::StorageClient,
 	OnlineClient, PolkadotConfig,
 };
 
+pub type HeaderStream = StreamOf<Result<(Header, BlockRef<H256>), subxt::Error>>;
+
+#[async_trait::async_trait]
+pub trait ApiClientT: DynClone + Send + Sync {
```
sandreim, line 41: Would we ever need to have two different implementations of this trait to easily switch between them ? If answer is no I think we don't really need a trait at the moment.

AndreiEres: I think we should keep both of them. The Light Client is slow, it takes some time to synchronize with the network. This can be annoying when we are looking for a historical data or need a quick run to get an immediate result.

Cargo.toml
AndreiEres: Maybe not the Light Client itself, but the integration of the Light Client with subxt will take some time to stabilize.

sandreim (approved):

PR #612, opened 2023-12-11
Use only one rpc client at time 

AndreiEres:
Changes have been made to improve the stability of the subxt client (including the light client mode). It's done in two specific parts.
- Updated subxt client implementation
- Improved rpc node requests

### Updated subxt client implementation
We used to spawn a client for a number of watched parachains or even more. Despite the memory and CPU usage it was a working approach until we decided to switch to a light client instead of a full RPC node. Spawning a light client for each parachain leads to overuse of resources and as a result - crashes the entire application.

What changed:
- Subxt wrapper replaced by frontend-backend realization of executor.
- Backend containing subxt client is initialized at startup.
- Frontend can be cloned between services without duplicating the subxt client.
- As a result, we have only one subxt client for the entire application.

### Improved rpc node requests
Implementing the new subxt client revealed a hidden problem with the number of requests. One client was unable to process all requests at the same speed as many clients and became a bottleneck. After investigation it was found that the most common request is to check the HRMP channels of a given parachain (about 80%).

We used to fetch the ingress and egress channel index of each parachain and then check each channel. It could be optimized because most of the attempts were empty, i.e. not every open channel was used to send a message in a given block.

What changed:
- To retreive the inbound channels, we moved from checking the inbound channel index to the HRMP channel digest. It shows which paras sent a message at the given block number for a given recipient. So we exclude empty attempts when fetching inbound channels.
- We joined requests whenever possible instead of awaiting them in a loop.
- In the meantime, we found a bug with incorrect channel id for outgoing messages: current para_id was set as receiver, but should be sender.

New implementation of HRMP channel requesting works well with a full RPC node, but light client can't hold that load. Since we currently only display HRMP metrics in CLI mode and don't send them to Prometheus, we decided to make them optional for a while. More work on improving HRMP fetching on the light client is coming.

#### Processing of HRMP channels for one block, while watching all parachains in the Polkadot network.

|        | Requests | Time |
|--------|----------|------|
| Before | 361 | 60.606s |
| After | 180  | 0.518s |

#### Memory usage, 5 minutes while watching all parachains in the Polkadot network.
|    | Memory, MB   | Graph |
|----|----------|-------|
| Before | 80.86 | old  |
| After, channels enabled | 27.92 | new |
| After, channels disabled | 26.45 | new_without_channels |


essentials/src/api/executor.rs
```diff
@@ -0,0 +1,495 @@
...
+	#[error("Send failed: {0}")]
+	PrioritySendError(#[from] PrioritySendError),
+	#[error("Recv failed: {0}")]
+	OneshotRecvError(#[from] OneshotRecvError),
+	#[error("Unexpected response for request {0:?}")]
+	UnexpectedResponse(Request),
+}
+
+impl RequestExecutorError {
+	pub fn should_repeat(&self) -> bool {
```
sandreim, line 122:
```suggestion
	pub fn should_retry(&self) -> bool {
```

essentials/src/api/executor.rs
```diff
@@ -0,0 +1,495 @@
...
+		api_client_mode: ApiClientMode,
+	) {
+		let client = match build_client(&url, api_client_mode, &self.retry).await {
+			Some(v) => v,
+			None => return,
+		};
+
+		loop {
+			tokio::select! {
+				message = from_frontend.recv() => {
```
sandreim, line 149: you don't really need a `select!` here, you can just `let message = from_frontend.recv().await`

essentials/src/api/executor.rs
```diff
@@ -0,0 +1,495 @@
...
+			)),
+			GetBestBlockSubscription => ChainSubscription(client.stream_best_block_headers().await?),
+			GetFinalizedBlockSubscription => ChainSubscription(client.stream_finalized_block_headers().await?),
+		};
+
+		Ok(response)
+	}
+}
+
+pub trait RequestExecutorNodes {
```
sandreim, line 247: I am a bit slow to understand what does the `String` represent?

sandreim: is it the rpc node url ?

essentials/src/api/executor.rs
```diff
@@ -0,0 +1,495 @@
...
+impl RequestExecutorBackend {
+	async fn run(
+		&mut self,
+		from_frontend: PriorityReceiver<ExecutorMessage>,
+		url: String,
+		api_client_mode: ApiClientMode,
+	) {
+		let client = match build_client(&url, api_client_mode, &self.retry).await {
+			Some(v) => v,
+			None => return,
```
sandreim, line 144: Shouldn't we log an error here ?

sandreim (approved):
Nice! Did you measure startup time and resource usage vs before this change ?

essentials/src/api/executor.rs
AndreiEres: Yes, It's a bit odd way to allow pass different types of node urls as an argument.

Issue #673, opened 2024-03-01
Async backing enabled chains break `introspector_pc_backed_count` metric


sandreim:
After collators started producing 6s blocks we can see the metrics are not right anymore.

https://grafana.teleport.parity.io/goto/Ngy4PeTSR?orgId=1

PR #675, opened 2024-03-05
Fix candidate tracking on async backing

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/673

Major changes
- Instead of a single current candidate, we track [included, backed] candidates for each assigned core.
- We set the `included` status based on chain events, which we store in the collector.
- Dead code removed

### DIFF
--- a/parachain-tracer/src/parachain_block_info.rs
+++ b/parachain-tracer/src/parachain_block_info.rs
@@ -48,11 +48,6 @@ impl ParachainBlockInfo {
 		BlakeTwo256::hash_of(&(&candidate.candidate.descriptor, commitments_hash))
 	}
 
-	#[cfg(test)]
-	pub fn set_backed(&mut self) {
-		self.state = ParachainBlockState::Backed
-	}
-
 	pub fn set_pending(&mut self) {
 		self.state = ParachainBlockState::PendingAvailability
 	}
@@ -96,12 +91,9 @@ mod tests {
 		assert!(!info.is_bitfield_propagation_slow());
 
 		info.max_availability_bits = 200;
-		assert!(!info.is_bitfield_propagation_slow());
+		assert!(info.is_bitfield_propagation_slow());
 
-		info.bitfield_count = 100;
-		assert!(!info.is_bitfield_propagation_slow());
-
-		info.set_backed();
+		info.bitfield_count = 120;
 		assert!(info.is_bitfield_propagation_slow());
 	}
 }
--- a/parachain-tracer/src/parachain_block_info.rs
+++ b/parachain-tracer/src/parachain_block_info.rs
@@ -35,7 +35,7 @@ pub struct ParachainBlockInfo {
 	/// Core occupation status.
 	pub core_occupied: bool,
 	/// The current state.
-	state: Vec<ParachainBlockState>,
+	state: ParachainBlockState,
 }
 
 impl ParachainBlockInfo {
@@ -49,40 +49,35 @@ impl ParachainBlockInfo {
 	}
 
 	pub fn set_idle(&mut self) {
-		self.state.push(ParachainBlockState::Idle)
+		self.state = ParachainBlockState::Idle
 	}
 
 	pub fn set_backed(&mut self) {
-		self.state.push(ParachainBlockState::Backed)
+		self.state = ParachainBlockState::Backed
 	}
 
 	pub fn set_pending(&mut self) {
-		self.state.push(ParachainBlockState::PendingAvailability)
+		self.state = ParachainBlockState::PendingAvailability
 	}
 
 	pub fn set_included(&mut self) {
-		self.state.push(ParachainBlockState::Included)
+		self.state = ParachainBlockState::Included
 	}
 
 	pub fn is_idle(&self) -> bool {
-		self.state.last() == Some(&ParachainBlockState::Idle)
+		self.state == ParachainBlockState::Idle
 	}
 
 	pub fn is_backed(&self) -> bool {
-		self.state.last() == Some(&ParachainBlockState::Backed)
+		self.state == ParachainBlockState::Backed
 	}
 
 	pub fn is_pending(&self) -> bool {
-		self.state.last() == Some(&ParachainBlockState::PendingAvailability)
+		self.state == ParachainBlockState::PendingAvailability
 	}
 
 	pub fn is_included(&self) -> bool {
-		self.state.last() == Some(&ParachainBlockState::Included)
-	}
-
-	pub fn has_state_changed(&self) -> bool {
-		let len = self.state.len();
-		len == 1 || len > 1 && self.state[len - 2] != self.state[len - 1]
+		self.state == ParachainBlockState::Included
 	}
 
 	pub fn is_bitfield_propagation_slow(&self) -> bool {
--- a/parachain-tracer/src/parachain_block_info.rs
+++ b/parachain-tracer/src/parachain_block_info.rs
@@ -89,10 +89,6 @@ impl ParachainBlockInfo {
 		len == 1 || len > 1 && self.state[len - 2] != self.state[len - 1]
 	}
 
-	pub fn is_data_available(&self) -> bool {
-		self.current_availability_bits > (self.max_availability_bits / 3) * 2
-	}
-
 	pub fn is_bitfield_propagation_slow(&self) -> bool {
 		self.max_availability_bits > 0 && !self.is_idle() && self.bitfield_count <= (self.max_availability_bits / 3) * 2
 	}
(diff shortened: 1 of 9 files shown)


parachain-tracer/src/parachain_block_info.rs
```diff
@@ -115,51 +108,10 @@ enum ParachainBlockState {
 mod tests {
 	use crate::test_utils::create_para_block_info;
 
-	#[test]
```
sandreim, line 118: Tests need to be updated, not removed.

parachain-tracer/src/parachain_block_info.rs
```diff
@@ -38,58 +34,55 @@ pub struct ParachainBlockInfo {
 	pub assigned_core: Option<u32>,
 	/// Core occupation status.
 	pub core_occupied: bool,
+	/// The current state.
+	state: Vec<ParachainBlockState>,
```
sandreim, line 38: Why more objects of state instead of one ?

parachain-tracer/src/parachain_block_info.rs
AndreiEres: The method it tested doesn't exist anymore.

parachain-tracer/src/tracker.rs
```diff
@@ -423,31 +475,36 @@ impl SubxtTracker {
...
-		if self.current_candidate.is_idle() {
-			progress.events.push(ParachainConsensusEvent::SkippedSlot);
-			stats.on_skipped_slot(progress);
-			metrics.on_skipped_slot(progress);
-		}
+		for candidate in self.all_candidates() {
+			if candidate.is_idle() {
+				progress.events.push(ParachainConsensusEvent::SkippedSlot);
+				stats.on_skipped_slot(progress);
+				metrics.on_skipped_slot(progress);
```
sandreim, line 482: insert `continue` . we don't need to check next conditions.

parachain-tracer/src/tracker.rs
```diff
@@ -423,31 +475,36 @@ impl SubxtTracker {
...
+			}
 
-		if self.current_candidate.is_backed() {
-			if let Some(candidate_hash) = self.current_candidate.candidate_hash {
-				progress.events.push(ParachainConsensusEvent::Backed(candidate_hash));
-				stats.on_backed();
-				metrics.on_backed(self.para_id);
+			if candidate.is_backed() {
```
sandreim, line 485: There seem to be so many ifs here, a match would be better if we had an enum for the candidate state.

parachain-tracer/src/tracker.rs
```diff
@@ -497,25 +563,31 @@ impl SubxtTracker {
...
+	}
+
 	fn is_just_backed(&self) -> bool {
 		self.last_backed_at_block_number.is_some() &&
 			self.last_backed_at_block_number == self.current_relay_block.map(|v| v.num)
 	}
 
-	fn is_slow_availability(&self) -> bool {
-		self.current_candidate.core_occupied &&
+	fn is_slow_availability(&self, core: u32) -> bool {
```
sandreim, line 584: I think it is a good time to make sure we have tests covering all of these functions and ensure they work as expected. The condition below seems brittle.

parachain-tracer/src/tracker.rs
```diff
@@ -299,21 +349,18 @@ impl SubxtTracker {
...
-				} else if !self.is_just_backed() {
-					self.current_candidate.set_pending();
+		let core_ids: Vec<u32> = self.cores.keys().cloned().collect();
+		for core in core_ids {
+			if self.is_current_candidate_backed(core) {
+				let is_just_backed = self.is_just_backed();
+				let max_bits = self.validators_indices(block_hash, storage).await?.len() as u32;
+				let candidate = self.current_candidate_mut(core).expect("Checked above");
+
+				candidate.current_availability_bits = extract_availability_bits_count(&bitfields, core);
```
sandreim, line 359: If the candidate was backed at current block, the bits should not be set at all. I'd assume we want to do this only if `!is_just_backed`, below.

parachain-tracer/src/tracker.rs
AndreiEres: It's covered here
https://github.com/paritytech/polkadot-introspector/blob/ae52a2c72e92d920b8d4b84d15974219e3ae8ce9/parachain-tracer/src/tracker.rs#L1298-L1319

parachain-tracer/src/tracker.rs
```diff
@@ -232,38 +241,79 @@ impl SubxtTracker {
...
+				if let Some(stored_candidate) = storage.candidate(candidate_hash).await {
+					if stored_candidate.is_included() {
+						candidate.set_included();
+						last_included = candidate.candidate_hash;
+					}
+				}
+			}
+		}
+		if last_included.is_some() {
+			self.relay_forks.last_mut().expect("must have relay fork").included_candidate = last_included;
```
sandreim, line 262: Can this expect panic ?

parachain-tracer/src/tracker.rs
```diff
@@ -232,38 +241,79 @@ impl SubxtTracker {
...
+		let mut last_included = None;
+		for candidate in self.all_candidates_mut() {
+			if let Some(candidate_hash) = candidate.candidate_hash {
+				if let Some(stored_candidate) = storage.candidate(candidate_hash).await {
+					if stored_candidate.is_included() {
+						candidate.set_included();
+						last_included = candidate.candidate_hash;
+					}
+				}
+			}
```
sandreim, line 259:
```suggestion
			let Some(candidate_hash) = candidate.candidate_hash  else {
			      continue
			}
		       let Some(stored_candidate) = storage.candidate(candidate_hash).await else {
		              continue
		        }
			if stored_candidate.is_included() {
				candidate.set_included();
				last_included = candidate.candidate_hash;
			}

```

parachain-tracer/src/tracker.rs
```diff
@@ -232,38 +241,79 @@ impl SubxtTracker {
...
-			self.current_candidate.set_candidate(candidate);
-			self.last_backed_at_block_number = Some(block_number);
-
-			if let Some(current_fork) = self.relay_forks.last_mut() {
-				current_fork.backed_candidate = self.current_candidate.candidate_hash;
+		let candidate_hashes = backed_candidates_by_para_id(backed_candidates, self.para_id)
+			.map(|v| ParachainBlockInfo::candidate_hash(&v));
+		let mut used_cores = vec![];
+		for candidate_hash in candidate_hashes {
+			if let Some(core) = self.candidate_core(candidate_hash, storage).await {
```
sandreim, line 279: use `let Some(..)` to avoid nesting.

parachain-tracer/src/tracker.rs
```diff
@@ -423,51 +471,48 @@ impl SubxtTracker {
...
-				metrics.on_slow_availability(self.para_id);
+		for candidate in self.all_candidates() {
+			match candidate.state {
+				ParachainBlockState::Idle => {
+					progress.events.push(ParachainConsensusEvent::SkippedSlot);
+					stats.on_skipped_slot(progress);
+					metrics.on_skipped_slot(progress);
+				},
+				ParachainBlockState::Backed =>
+					if let Some(candidate_hash) = candidate.candidate_hash {
```
sandreim, line 482: Could we remove the option ? Doesn't seem to make sense to have the state backed but no hash.

parachain-tracer/src/tracker.rs
```diff
@@ -423,51 +471,48 @@ impl SubxtTracker {
...
+						progress.events.push(ParachainConsensusEvent::Backed(candidate_hash));
+						stats.on_backed();
+						metrics.on_backed(self.para_id);
+					},
+				ParachainBlockState::PendingAvailability => {
+					if self.is_slow_availability(candidate.assigned_core.expect("must be assigned")) {
```
sandreim, line 488: When would expect panic ? I'd avoid doing this since it reduces the uptime of the monitor. Let's log errors instead.

parachain-tracer/src/tracker.rs
AndreiEres: Above we match `ParachainBlockState::Idle` where candidate has no hash.

parachain-tracer/src/tracker.rs
AndreiEres: It should never panic, because a backed candidate info should include assigned core.

AndreiEres:
Updated, removing `Idle` state from candidate's state. Used `None` and `Some(block_info)`. So we removed unnecessary `.expect()`s.

sandreim (approved):

Issue #769, opened 2024-08-19
Crashing on Rococo Chain


lazam:
version: v0.2.27-metadata-v7
tool: polkadot-parachain-tracer

```
2024-08-19 12:25:56.467	-----------------------------------------------------------------------
2024-08-19 12:26:11.135	[2024-08-19T10:26:11.135376Z ERROR polkadot_introspector_essentials::chain_head_subscription] Subscription to wss://rococo-rpc.polkadot.io:443 failed, received None instead of an event
2024-08-19 12:26:27.583	Parachain Tracer
2024-08-19 12:26:27.583		will trace all parachain(s)
2024-08-19 12:26:27.583		on wss://rococo-rpc.polkadot.io:443
2024-08-19 12:26:27.583		using RPC Client
2024-08-19 12:26:27.583	
2024-08-19 12:26:43.188	Host configuration for wss://rococo-rpc.polkadot.io:443:
2024-08-19 12:26:43.188		👀 Max validators: 200 / 0 per core
2024-08-19 12:26:43.188		👍 Needed approvals: 5
2024-08-19 12:26:43.188		🥔 No show slots: 2
2024-08-19 12:26:43.188		⏳ Delay tranches: 89
2024-08-19 12:26:43.188	-----------------------------------------------------------------------
2024-08-19 12:29:33.543	[2024-08-19T10:29:33.543584Z ERROR polkadot_introspector_essentials::api::executor] Cannot process the request GetClaimQueue(0x52454b0c57c60d709e9dd4f830553d250804c5b5b3df04253b3605974fb9826b): DynamicError(DecodeDynamicError("v7 assignment", Variant(Variant { name: "Pool", values: Named([("para_id", Value { value: Composite(Unnamed([Value { value: Primitive(U128(4484)), context: 4 }])), context: 152 }), ("core_index", Value { value: Composite(Unnamed([Value { value: Primitive(U128(5)), context: 4 }])), context: 444 })]) })))
2024-08-19 12:29:33.543	[2024-08-19T10:29:33.543656Z ERROR polkadot_introspector_essentials::collector] collector service could not process event: ExecutorFatal(OneshotRecvError(RecvError(())))
2024-08-19 12:29:33.544	[2024-08-19T10:29:33.544874Z ERROR polkadot_parachain_tracer] Shutting down, Collector's executor error: Recv failed: channel closed
2024-08-19 12:30:01.619	Parachain Tracer
[...]

PR #770, opened 2024-08-22
[parachain-tracer] Introduce top level restart

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/769

Introduces two types of shutdown: 
1. Graceful – when we stop the application with ctrl-c.
2. Restart – something bad happened, generaly lost the connection to RPC node.

On restart we reinitialize the application from the beginning. Same rules for retry as in other parts of application applied. 

Additionally in the other PR was updated subxt, which added own reconnection to RPC-nodes.

### DIFF
--- a/parachain-tracer/src/main.rs
+++ b/parachain-tracer/src/main.rs
@@ -37,13 +37,12 @@ use polkadot_introspector_essentials::{
 	api::{api_client::ApiClientMode, executor::RequestExecutor},
 	chain_head_subscription::ChainHeadSubscription,
 	chain_subscription::ChainSubscriptionEvent,
-	collector,
-	collector::{Collector, CollectorOptions, CollectorStorageApi, CollectorUpdateEvent, TerminationReason},
+	collector::{self, Collector, CollectorOptions, CollectorStorageApi, CollectorUpdateEvent, TerminationReason},
 	consumer::{EventConsumerInit, EventStream},
 	historical_subscription::HistoricalSubscription,
-	init,
+	init::{self, Shutdown},
 	types::BlockNumber,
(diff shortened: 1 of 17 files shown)


s0me0ne-unkn0wn:
That sounds like healing symptoms instead of healing the disease, doesn't it? Is there any objective reason why the root cause cannot be fixed?

sandreim:
> That sounds like healing symptoms instead of healing the disease, doesn't it? Is there any objective reason why the root cause cannot be fixed?

Yes indeed. Do we know why it happens? Maybe something else needs a fix.

AndreiEres:
RPC nodeы drop connections after a while or can be unreachable. But I agree, that it's too straightforward.

alexggh (approved):
Looks like an improvement to the current state.

However, I do second the previous comments that restarting on errors might mask other bugs.

parachain-tracer/src/main.rs
```diff
@@ -312,9 +317,10 @@ impl ParachainTracer {
 									trackers.len(), futures.len());
 								match reason {
 									TerminationReason::Normal => break,
-									TerminationReason::Abnormal(code, info) => {
+									TerminationReason::Abnormal( info) => {
```
s0me0ne-unkn0wn, line 320:
```suggestion
									TerminationReason::Abnormal(info) => {
```

s0me0ne-unkn0wn (approved):
Looks good as a solution for the current problem :+1: RPC is evil anyway, let's just hope we'll get rid of it altogether at some point.

sandreim (approved):

PR #864, opened 2025-05-27
Update subxt

AndreiEres:


essentials/src/api/mod.rs
```diff
@@ -142,9 +148,10 @@ mod tests {
 		let api =
 			ApiService::<H256>::new_with_storage(RecordsStorageConfig { max_blocks: 1 }, request_executor().await);
 		let mut subxt = api.executor();
+		let hasher = subxt.hasher(rpc_node_url()).unwrap();
 
 		let head = subxt.get_block_head(rpc_node_url(), None).await.unwrap().unwrap();
-		let cores = subxt.get_occupied_cores(rpc_node_url(), head.hash()).await;
+		let cores = subxt.get_occupied_cores(rpc_node_url(), hasher.hash_of(&head)).await;
```
jsdw, line 154: I wonder whether you can avoid needing to expose the `hasher` for many of these calls by doing one of the following:

1. Have a function like `get_block_hash` rather (or as well as) `get_block_head`. In Subxt, blocks have [a `.hash()` method](https://docs.rs/subxt/latest/subxt/blocks/struct.Block.html#method.hash) for instance that could return this.
2. Have functions accept a block header rather than a block hash (so that the hashing can be handled "internally" by Subxt). Perhaps this is difficult because it looks like you pass things through a channel, so would need owned values and headers are larger than hashes to have to be sending etc.
3. Functions could also accept something like a `trait IntoBlockHash { fn hash() }` type trait and then we have impls for hashes but also blocks or whatever. Maybe suffers from the same issue as 2 does.

Issue #810, opened 2024-11-04
disputes: Add metric label for validator address


sandreim:
Currently the validator index -> address mapping is only available in CLI mode, it would be useful to expose this as a label value in metrics for monitoring.

PR #887, opened 2025-06-30
Track dispute initiators

AndreiEres:
Fixes https://github.com/paritytech/polkadot-introspector/issues/810

### DIFF
--- a/parachain-tracer/src/prometheus.rs
+++ b/parachain-tracer/src/prometheus.rs
@@ -262,14 +262,24 @@ impl PrometheusMetrics for Metrics {
 				metrics
 					.disputes_stats
 					.initiators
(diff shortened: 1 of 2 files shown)


parachain-tracer/src/prometheus.rs
```diff
@@ -253,6 +257,21 @@ impl PrometheusMetrics for Metrics {
 					.with_label_values(&[&para_string])
 					.inc();
 			}
+
+			for (index, address) in dispute_outcome.initiators.iter() {
+				metrics
+					.disputes_stats
+					.initiators
+					.with_label_values(&[&para_string, &index.to_string(), address])
```
alexggh, line 265: I think we need the session here as well, otherwise we will really get confused, because the validator Index is relevant only in the same session.

PR #943, opened 2026-03-18
Fix parachain-tracer historical mode exit and Ctrl+C handling

AndreiEres:
Fixes historical mode hanging after processing all blocks by sending a Termination event instead of entering an infinite heartbeat loop


AndreiEres:
Mostly changes here — adding indents, but GH diff doesn't catch it. Tip: look at the diff in vscode.

alexggh (approved):

PR #966, opened 2026-07-02
Decode chain reads without metadata, behind a shadow flag

AndreiEres:
## What this is

Every chain read that parachain-tracer, block-time and whois make through subxt's metadata decoding gets a second implementation that needs no metadata. `subxt` stays as the transport. Only its decoding role is replaced. Nothing is switched over: the metadata path still returns every value.

## Why

The checked-in `.scale` file needs a CI refresh twice a day, and every new metadata version drags in a subxt upgrade whose breaking changes churn our code. Decoding positionally from runtime calls and hand-built storage keys removes both.

## Where the data comes from instead

| Today | Without metadata |
|---|---|
| `System.Events` scrape (candidates, disputes) | `ParachainHost_candidate_events`, `ParachainHost_disputes` |
| typed and dynamic runtime APIs (cores, claim queue, groups, session index) | the same `ParachainHost_*` calls via `state_call` |
| `Babe.*` storage | `BabeApi_current_epoch`, plus a manual key for the per-block slot |
| typed storage (timestamp, session keys, HRMP) | hand-built keys, hashers checked against polkadot-sdk |
| subxt blocks API (head, number, bitfields) | `chain_getHeader`, `chain_getBlock` |

Decoders live in `essentials/src/api/decode.rs`. They read only the leading fields we use, positionally, so a field appended in a runtime upgrade is tolerated and anything else fails loud rather than returning a wrong value.

## How it is verified

`--shadow-decode-without-metadata` decodes each migrated read twice at the same block and compares. A mismatch aborts the process, reporting the block, the read and both values. With the flag off, nothing changes.

A coverage checklist tells "ran clean" from "never ran", so a quiet run cannot pass for a proven one. All 20 shadowed reads ran clean on Polkadot, Kusama and Westend, no mismatches. A Westend pod has run the shadow build since the end of August and restarts as often as the pods on the normal build.

## Two features dropped

The host-configuration startup banner, and on-demand order tracking with its `pc_para_on_demand` metrics. Both read structures that no runtime call exposes, and the on-demand delay metrics would otherwise report a permanent zero.

## What comes next

Run the shadow build on the Kusama and Polkadot pods and watch it there. Once it stays clean, the metadata-free path becomes the only one, and the subxt macro, the `.scale` asset and its CI refresh job go with it.

## Reviewing

Start with `decode.rs` and `shadow.rs`. `api_client.rs` is mechanical: each shadowed method fetches both values and compares them. The commits are one read each and read well in order.


essentials/src/collector/mod.rs
```diff
@@ -378,11 +378,98 @@ impl Collector {
...
+
+		let disputes = self.executor.get_disputes(self.endpoint.as_str(), hash).await?;
+		let metadata_free_initiated: Vec<H256> = disputes
+			.iter()
+			.filter(|dispute| dispute.start == block_number)
+			.map(|dispute| dispute.candidate_hash)
+			.collect();
+		let metadata_free_concluded: Vec<ConcludedDispute> = disputes
+			.iter()
+			.filter(|dispute| dispute.concluded_at == Some(block_number))
```
DenzelPenzel: I am not sure this is valid case in the real network or not. But can we rely on ```concluded_at``` here? If the other side reaches supermajority in a later block, it emits another ```DisputeConcluded``` but leaves ```concluded_at``` unchanged, so this filter misses the event. I reproduced this in a pallet test, though not on a running network

essentials/src/api/shadow.rs
```diff
@@ -0,0 +1,205 @@
...
+	if !old.is_empty() || !new.is_empty() {
+		mark(read);
+	}
+}
+
+fn report_and_abort(block: H256, read: &str, old: &str, new: &str) -> ! {
+	error!(
+		"shadow-decode mismatch at block {block:?} for read `{read}`:\n  metadata      = {old}\n  metadata-free = {new}"
+	);
+	std::process::exit(1);
```
DenzelPenzel, line 163: Could we return an error here and let the caller terminate?

parachain-tracer/src/tracker.rs
```diff
@@ -364,10 +338,7 @@ impl SubxtTracker {
 	}
 
 	async fn set_core_assignment(&mut self, block_hash: H256, storage: &TrackerStorage) {
-		for (core, scheduled_ids) in self.cores.clone() {
-			self.is_on_demand_scheduled_in_current_block =
-				self.on_demand_order.is_some() && scheduled_ids[0] == self.para_id;
-
+		for core in self.cores.clone().into_keys() {
```
DenzelPenzel: could we collect ```self.cores.keys().copied()``` instead of cloning the map and all its values?

essentials/src/api/api_client.rs
```diff
@@ -71,13 +89,22 @@ where
...
+	shadow: bool,
 }
 
 impl<T: OnlineClientT<PolkadotConfig>> ApiClient<T> {
 	pub fn hasher(&self) -> PolkadotHasher {
 		self.hasher
 	}
 
+	/// Whether the shadow-decode harness is enabled for this client.
+	pub fn shadow_enabled(&self) -> bool {
```
DenzelPenzel: is this getter needed?

essentials/src/collector/mod.rs
AndreiEres: good catch

essentials/src/api/shadow.rs
AndreiEres: Keeping the abort, there is a long story with the exiting the tool

parachain-tracer/src/tracker.rs
AndreiEres: right

essentials/src/api/api_client.rs
AndreiEres: dead code

DenzelPenzel (approved):

alexggh (approved):

